PR #45, opened 2015-11-06
Move Makefile-dist into Makefile; update documentation.

markro49:
Combine Makefile-dist and Makefile into a single Makefile.  Make related doumentation changes and improve sections on installing Daikon.


doc/daikon.texinfo
```diff
@@ -260,7 +260,7 @@ There are two additional requirements.  You must have a Java 7 (or later) JVM
 The @file{tools.jar} file that comes with your JVM must also be on your
 classpath.
 
-@xref{Installing Daikon}, for situations where you should
+See bullet point two above (@ref{Installing Daikon}), for situations where you should
 follow the complete installation instructions of
 @ref{Complete installation}.  (Also, if you do not know how to add a jar file to your
```
mernst: Isn't this documentation now obsolete, along with anything else that mentions the complete installation as a contrast to other approaches?

markro49: I left in the Simple Installation as it does work as described in the document.  Do you think it should be removed?

mernst: I see.  OK.

Makefile
```diff
@@ -139,7 +142,22 @@ TMPDIR ?= $(if $(shell if [ -d /scratch ] ; then echo true; fi),/scratch/$(USER)
 
 ### Default tag
 help:
-	@echo "Targets:"
+	@echo
+	@echo "This is the primary Makefile for both an archive file based setup"
+	@echo "(i.e. created from daikon-<version>.zip) and for a repository"
+	@echo "based setup (i.e. git cloned from //github.com/codespecs/daikon)."
+	@echo "This is primarily for users who wish to customize or extend DAIKON."
```
mernst: It's also for anyone who wants to use any tools together with Daikon, such as DynComp, right?
Can this paragraph be cut?

Makefile
```diff
@@ -139,7 +142,22 @@ TMPDIR ?= $(if $(shell if [ -d /scratch ] ; then echo true; fi),/scratch/$(USER)
...
+	@echo
+	@echo "These first two targets are only for the archive version:"
+	@echo " install                  -- build additional tools and kvasir"
```
mernst: In the Linux world, "make install" means to place the program's binary somewhere on your path.  This seems different than that, which I think will be confusing to users.

Makefile
```diff
@@ -139,7 +142,22 @@ TMPDIR ?= $(if $(shell if [ -d /scratch ] ; then echo true; fi),/scratch/$(USER)
...
+	@echo " install-test             -- simple verification of installtion"
```
mernst: Typo "installtion"

Makefile
```diff
@@ -139,7 +142,22 @@ TMPDIR ?= $(if $(shell if [ -d /scratch ] ; then echo true; fi),/scratch/$(USER)
...
+	@echo
+	@echo "This target is shared; it is for building a modified version"
+	@echo "of the java runtimes needed by the DynComp tool:"
```
mernst: "Java" should be capitalized.

Makefile
```diff
@@ -139,7 +142,22 @@ TMPDIR ?= $(if $(shell if [ -d /scratch ] ; then echo true; fi),/scratch/$(USER)
...
+	@echo " dyncomp-jdk              -- Make file java/dcomp_rt.jar"
+	@echo
+	@echo "The rest of the targets listed below are for a full"
+	@echo "repository based installation. For building:"
```
mernst: These are also relevant to an archive-based installation if one has made any changes, right?
It might be better to focus more on the use cases.  (I do agree that users generally shouldn't make changes after installing from the archive, but I'm not sure this message is the right place for that information.)

Makefile
```diff
@@ -90,8 +90,11 @@ INV_DIR := $(shell pwd)
 ifeq (cygwin,$(OSTYPE))
 #JAVA tools need Windows path on Windows
 JAR_DIR := $(shell cygpath -m $(INV_DIR))
+# for install-test target
+QT_PATH:=`cygpath -wp ../../../daikon.jar:.`
```
mernst: It's inconsistent that line 92 uses $(shell ...) and this line uses `...`.  I would make the two consistent.

Makefile
```diff
@@ -149,16 +153,39 @@ help:
 	@echo " kvasir                   -- make Kvasir, the C front end"
 	@echo " very-clean               -- remove (most) all generated files"
 	@echo " dyncomp-jdk              -- Make file java/dcomp_rt.jar"
-	@echo "Creating the Daikon distribution:"
+	@echo " quick-test               -- simple verification of system"
```
mernst: Is "of system" needed?

Makefile
```diff
@@ -139,6 +142,7 @@ TMPDIR ?= $(if $(shell if [ -d /scratch ] ; then echo true; fi),/scratch/$(USER)
 
 ### Default tag
 help:
+	@echo
```
mernst: What is the need for a blank line here?  The text is already quite long.

Makefile
mernst: Should this be next to the "junit" and "test" targets?  More generally, what order are the targets in?  Also explain that these are quick system tests; that isn't clear from the current description.

Makefile
```diff
@@ -149,16 +153,39 @@ help:
...
 	@echo " daikon.tar daikon.jar    -- just makes the tar files"
 	@echo " staging                  -- moves all release file to staging-daikon/"
 	@echo " test-staged-dist         -- tests the distribution in staging-daikon/"
 	@echo " staging-to-www           -- copies staging-daikon/ to website"
 	@echo " "
-	@echo "This Makefile is for manipulating the entire Daikon system."
 	@echo "Daikon source code is in the java/daikon subdirectory (but you"
 	@echo "can perform basic operations like compiling it from here)."
 
+distribution-check:
```
mernst: Can you add brief documentation of the purpose of each of these targets?

mernst: Shouldn't this be located in the file near the distribution targets?  And quick-test near the test targets?

Makefile
```diff
@@ -206,7 +233,9 @@ fjalar/valgrind/Makefile.am: ../fjalar/auto-everything.sh
 	ln -nsf ../fjalar fjalar
 	touch $@
 
-fjalar/valgrind/Makefile.in: fjalar/valgrind/Makefile.am
```
mernst: What is the purpose of this change?  I would have thought the before and after versions behaved the same (but the before seems more concise).

Makefile
```diff
@@ -533,6 +584,8 @@ daikon.tar daikon.zip: doc-all $(DOC_PATHS) $(EDG_FILES) $(README_PATHS) $(DAIKO
 	mkdir ${TMPDIR}/daikon
 
 	mkdir ${TMPDIR}/daikon/doc
+	# dummy www directory to remove end user Make warnings
```
mernst: Should we include the www directory in the release instead?  It's not very big, and doing so would eliminate a difference between the repository and the release.

doc/developer.texinfo
```diff
@@ -230,18 +230,19 @@ compile Daikon, you need a Java compiler such as @samp{javac} on your path.
...
 To override the default Java compiler (@samp{javac}), create a
-@file{Makefile.user} file in the @file{daikon/} directory and add a line
+@file{Makefile.user} file in the @file{daikon} (@env{$DAIKONDIR}) directory and add a line
 like the following.
 @example
 JAVAC ?= jikes -g +E +F
 @end example
 
-In order to compile Daikon, you need the C preprocessor
+In order to compile Daikon, you will also need the C preprocessor
```
mernst: I wouldn't include "will"; it's needed, rather than a future need.

doc/developer.texinfo
```diff
@@ -2026,7 +2028,15 @@ will take at least two hours --- but could be much more depending on the number
 different platforms on which you test the release.
 
 @noindent
-NOTE: The default version of the Java JDK on the majority of the CSE
+@c The following few lines are a texinfo hack to get IMPORTANT in hightlighted font.
```
mernst: I don't understand the goal, and the use of fake URLs is weird to me.  I don't think we should be using `@ref`, `@anchor`, and such except for cross-references.
Why not just use `@strong` if you want emphasis?  (Same comment elsewhere.)

Makefile
```diff
@@ -89,9 +89,12 @@ INV_DIR := $(shell pwd)
 
 ifeq (cygwin,$(OSTYPE))
 #JAVA tools need Windows path on Windows
-JAR_DIR := $(shell cygpath -m $(INV_DIR))
+JAR_DIR :=`cygpath -wp $(INV_DIR)`
```
mernst: As a minor issue, the space before and after the := lines is inconsistent.  Same two lines down.

PR #48, opened 2015-12-14
Updates to release procedure

markro49:
Addresses all our proposed changes except:
- moving staging-daikon up a level
- combining steps 19-22 (which have been renumbered)


Makefile
```diff
@@ -448,25 +448,27 @@ staging: doc/CHANGES
 	# Delete the tmp files
 	cd ${TMPDIR} && /bin/rm -rf daikon daikon.dist old_tar.txt new_tar.txt
 
-# Copy the files in the staging area to the website.  This will copy
```
mernst: This comment is a bit confusing because the staging area is on the website (it's served by the webserver).  I think you mean to the live or main website.

Makefile
```diff
@@ -448,25 +448,27 @@ staging: doc/CHANGES
...
-# all of the files in staging, but will not delete any files in the website
-# that are not in staging.
-staging-to-www: $(STAGING_DIR)
-#copy the files
+# Copy the files in the staging area to the website.
+# We do this by deleting the previous release directory and
+# renaming the staging directory to be the release directory.
+staging-to-www: $(STAGING_DIR) save-current-release
+	@echo "staging-to-www is disabled while we test new layout."
```
mernst: Is this necessary?  We can just be careful not to run it until we are happy with it.

Makefile
```diff
@@ -448,25 +448,27 @@ staging: doc/CHANGES
...
+	\mv $(WWW_DIR)/download/cvs $(STAGING_DIR)/download/cvs
+# Remove previous release
+	\rm -rf $(WWW_DIR)
+# Rename staging
+	\mv $(STAGING_DIR) $(WWW_DIR)
 # This command updates the dates and sizes in the various index files
-	html-update-link-dates $(DIST_DIR)/index.html
-# with new version number system, this is now done manually
-#	$(MAKE) update-dist-version-file
+	html-update-link-dates $(WWW_DIR)/download/index.html
```
mernst: Shouldn't this have been done by "make staging"?  I would think that the staging-to-www job would do nothing but copy files, without changing any contents.  (And, this command shouldn't end up making any changes anyway, should it?)

Makefile
```diff
@@ -448,25 +448,27 @@ staging: doc/CHANGES
...
+# and we're done
+	-chmod -R u-w $(WWW_DIR)
+	-chmod u-w $(WWW_PARENT)
 	@echo "*****"
 	@echo "Don't forget to send mail to daikon-announce and commit changes."
 	@echo "(See sample messages in ~mernst/research/invariants/mail/daikon-lists.mail.)"
```
mernst: Change this to indicate that the sample message is now in the developer manual.

doc/developer.texinfo
```diff
@@ -2446,27 +2436,150 @@ GNU Texinfo Manual (@uref{http://www.gnu.org/software/texinfo/manual/texinfo/tex
...
+
+[Time: moments]
+
+
+@need 1000
+@item
+It is important to build the release candidate from a clean enlistment.
+This is to ensure we are testing only the files intended for release and
+that those files are the same as a user would receive with a clone of
+the repository. It is also recommend doing this step on your system's
```
mernst: I would drop the last two sentences.  A reader should just follow your instructions, and so it will naturally be in /tmp/$USER.

doc/developer.texinfo
```diff
@@ -2446,27 +2436,150 @@ GNU Texinfo Manual (@uref{http://www.gnu.org/software/texinfo/manual/texinfo/tex
...
+Build everything to be distributed and copy it to the temporary staging area
+@c not @uref{...} because we don't want robots to notice and crawl the link
+(http://plse.cs.washington.edu/staging-daikon/).
+
+@example
 make staging
 @end example
 
+@noindent
+If the command immediately stops with the error message:
```
mernst: This won't happen any more because an earlier step will check doc/CHANGES and the "make staging" step won't.

doc/developer.texinfo
```diff
@@ -2446,27 +2436,150 @@ GNU Texinfo Manual (@uref{http://www.gnu.org/software/texinfo/manual/texinfo/tex
...
+cd fjalar
+git commit -a -m "Change version to $DAIKONVERSION"
+git push
+@end example
+
+[Time: moments]
+
+
+@need 1000
+@item
```
mernst: I suggest that the next three steps be combined into just one step in the manual.  It can start with creating the new command shell, but then all the commands can just be one block that can be pasted as a whole.  When you tell the user to start over, then there is just one command to redo, making it less confusing where to re-start.

doc/developer.texinfo
```diff
@@ -2446,27 +2436,150 @@ GNU Texinfo Manual (@uref{http://www.gnu.org/software/texinfo/manual/texinfo/tex
...
+cd /tmp/$USER/daikon
+export DAIKONDIR=`pwd`
+export JAVA_HOME=/homes/gws/markro/jdk7
+source $DAIKONDIR/scripts/daikon.bashrc
+@end example
+
+[Time: moments]
+
+@noindent
+Each of the steps below assumes that the current working directory is
```
mernst: There is only one step below, so "each of the steps below" is confusing.  But, I suggest that you consolidate multiple steps into one as suggested above.

doc/developer.texinfo
```diff
@@ -2446,27 +2436,150 @@ GNU Texinfo Manual (@uref{http://www.gnu.org/software/texinfo/manual/texinfo/tex
...
+@noindent
+You may simply:
+@example
+touch doc/CHANGES
+@end example
+
+The final output of this step
+will be a list of files that were added/removed since the
+last release.
+If any of these differences is unexpected, then investigate and correct the
```
mernst: When correcting it, I would suggest that be done in the main repository and committed, and the full "make staging" step (from starting a new command shell) redone.

doc/developer.texinfo
```diff
@@ -2446,27 +2436,150 @@ GNU Texinfo Manual (@uref{http://www.gnu.org/software/texinfo/manual/texinfo/tex
...
+
+@c Time on buffalo as of 12/16/2015
+@c [Time: 34 min]
+@c I've notified mernst about this 10x slow down.
 [Time: 4 min]
 
 
+@need 1000
 @item
 Check the website to see if it has any broken links.
```
mernst: I would say "check the staging website" rather than "the website".

Makefile
```diff
@@ -448,25 +448,27 @@ staging: doc/CHANGES
...
-# 	\rm -f /cse/web/research/plse/daikon/download/daikon-*
-# # don't trash existing history directory
-# #	(cd $(STAGING_DIR) && tar cf - --exclude=history .) | (cd $(WWW_DIR) && tar xfBp -)
-# Remove previous files, except history and staging-daikon
-	(cd $(STAGING_DIR) && \ls | grep -v history | grep -v staging-daikon | xargs \rm -rf
-# Copy directory
-	rsync -Cavz --quiet $(STAGING_DIR) $(WWW_DIR)
-#	-chmod -R u-w $(WWW_DIR)
+	-chmod -R u+w $(STAGING_DIR)
+# Move 'permanent' directories that were not included in staging
```
mernst: I suggest that these directories be copied (be sure to use -p to preserve timestamps) to the staging website, as part of "make staging".  That way the staging website is more complete, which is good to avoid confusion, to make links work, etc.  It's also shorter in the Makefile than the current scheme, since this target will get shorter than the "make staging" target will get longer.

doc/developer.texinfo
```diff
@@ -2502,15 +2615,28 @@ If the web search is not successful, email the maintainer of the old content.
...
 Note that you must fix it in the original repository,
-not in the drop directories.
-This means correcting the file, most likely located beneath either
-@file{$DAIKONDIR/doc} or @file{$BIBDIR}, committing the changes, and
-returning to the previous step (@command{make staging}) prior to re-running
-the broken link tool.
+not in the drop directories. This means:
+@itemize
+@item
+Switch back to your original Bash shell.
```
mernst: ... which has DAIKONDIR set to your main clone.

doc/developer.texinfo
```diff
@@ -2502,15 +2615,28 @@ If the web search is not successful, email the maintainer of the old content.
...
+@item
+Correct the file. (Most likely located beneath either
+@file{$DAIKONDIR/doc} or @file{$BIBDIR}.)
+@item
+Commit the changes.
+@item
```
mernst: I think it's simpler to do the whole "make staging" process (from creating the new command shell) all over again.  It might take a bit longer in wall clock time, but it's no more human time and likely less.

doc/developer.texinfo
```diff
@@ -2649,20 +2777,23 @@ make distribution-check
...
-[Time: moments]
+At this point we are done building and testing the release candidate and you
+should exit the Bash command shell created in step 12 and return to your
+original Bash shell.
 
 
 @need 1000
 @item
-Update the website.
+Update the website by moving the current release to the history
```
mernst: Does this step still move the current release to the history archives?  I thought that "make staging" also copied the release there (at least in its own copy of the history).

Makefile
markro49: changed to:
 Convert the staging area to the live release.

Makefile
markro49: I was worried about accidentally deleting to current release but I just made a backup copy instead.

Makefile
markro49: This section was changed with the next commit that moved the staging directory up and level.  The history and cvs directories are moved as part of staging-to-www not copied as part of staging.  Would you still prefer the latter?

Makefile
markro49: good idea, i will make the change

Makefile
markro49: done

doc/developer.texinfo
markro49: done

doc/developer.texinfo
markro49: done

doc/developer.texinfo
markro49: Current version does 'save-current-release' as part of 'staging-to-www'.  I don't think we need to save it earlier as part of 'staging'.

Makefile
mernst: I think it makes sense to copy (not move) the history and cvs directories as part of "make staging".  Then make staging-to-www can just do a clean move.

markro49: done

doc/developer.texinfo
markro49: I don't think we can just say repeat the step - won't a 'git pull' be required?

doc/developer.texinfo
markro49: correct

doc/developer.texinfo
markro49: removed that line

doc/developer.texinfo
markro49: same problem with needing a 'git pull'?

doc/developer.texinfo
mernst: If an early step is to delete any existing clone that is in the /tmp/... location, then "git pull" is not required.

doc/developer.texinfo
mernst: There is `rm -rf /tmp/$USER/daikon` right before the `git clone`, so `git clone` is not necessary.

doc/developer.texinfo
mernst: This makes the regular directory after "make staging-to-www" different than the staging directory before the command.  I think it would be better to put all the work of setting up the distribution website in one place rather than distributing it, partly in "make staging" and partly in "make staging-to-www".

doc/developer.texinfo
markro49: I'm confused.  In an earlier note you said to start combining the steps with the get new shell step - but this is one step after creating the clone.  So we need to go back to that step or combine that step as well or ?

doc/developer.texinfo
markro49: done

doc/developer.texinfo
mernst: I suggest that it be one step, starting with creating a new shell and doing all the work in the new shell.  The first step of that would be creating the temporary directory, deleting the daikon subdirectory, and cloning.

Makefile
```diff
@@ -390,23 +388,28 @@ repository-test:
 
 # Main distribution
 
+check-repo: doc/CHANGES
```
mernst: Please document what this does.  "check" is vague and easy to misinterpret.
Why does it depend on doc/CHANGES?  Would there be any harm in not depending on doc/CHANGES?
Also, is the result expected to be viewed by a human or to halt the process if it doesn't return a status of 0?  (I expect the latter.)

Makefile
```diff
@@ -390,23 +388,28 @@ repository-test:
...
 # to be the owner of an existing file, the permissions system gives
 # that priority and we must set the write bit.  These two chmod commands
 # will fail if you are not the owner, but the remainder of the commands
 # should work fine.
+	-chmod u+w $(WWW_PARENT)
 	-chmod -R u+w $(STAGING_DIR)
-	-chmod u+w $(WWW_DIR)
 	/bin/rm -rf $(STAGING_DIR)
-	# dummy history directory to remove checklink warnings
 	install -d $(STAGING_DIR)/history
```
mernst: Why not copy the history directory, so that the staging and real directories are as similar as possible?

Makefile
```diff
@@ -448,28 +455,23 @@ staging: doc/CHANGES
...
-# Remove previous files, except history and staging-daikon
-	(cd $(STAGING_DIR) && \ls | grep -v history | grep -v staging-daikon | xargs \rm -rf
-# Copy directory
-	rsync -Cavz --quiet $(STAGING_DIR) $(WWW_DIR)
-#	-chmod -R u-w $(WWW_DIR)
-# This command updates the dates and sizes in the various index files
-	html-update-link-dates $(DIST_DIR)/index.html
-# with new version number system, this is now done manually
-#	$(MAKE) update-dist-version-file
+	-chmod -R u+w $(STAGING_DIR)
```
mernst: Since you are just moving the staging directory, not changing any of its contents, I don't think you need to do anything to its permissions.

Makefile
```diff
@@ -448,28 +455,23 @@ staging: doc/CHANGES
...
+	# Remove previous release
```
mernst: I would drop the comments here given that you have described the process in the comments at lines 459-460.

Makefile
```diff
@@ -494,31 +496,37 @@ doc-all-except-pdf:
...
 
 # Perl command compresses multiple spaces to one, for first 9 days of month.
 ifeq ($(origin TODAY), undefined)
 TODAY := $(shell date "+%B %e, %Y" | perl -p -e 's/  / /')
 endif
 
+update-and-commit-version: update-doc-dist-date-and-version
+	git commit -a -m "Change version to $(NEW_VER)"
+	git push
+	cd fjalar && git commit -a -m "Change version to $(NEW_VER)" && git push
```
mernst: If you don't want to change the working directory, I think you need to enclose this line in parentheses.  (You would need to do that in the shell and I think in a Makefile too.)

doc/developer.texinfo
```diff
@@ -2446,34 +2436,99 @@ GNU Texinfo Manual (@uref{http://www.gnu.org/software/texinfo/manual/texinfo/tex
...
-any failures, then investigate and correct as necessary.
-The final output of this step
-will be a list of files that were added/removed since the
-last release.
-If any of these differences is unexpected, then investigate and correct the
-difference, and redo the step.
+Once you have successfully reached this point, you are ready to make a release
+candidate and begin testing it. Any source changes required to fix a problem
+encountered above must be committed to the repository now.
+Verify doc/CHANGES is up-to-date and there are no uncommitted files.
```
mernst: It's not clear that this is a description of the steps that follow.  I thought it was something that I had to do now, and I wondered how to do it.  This may just be an issue with the way the instructions are written; I was expecting a list of steps to perform rather than a description of what will come, followed by a list of steps to perform.  maybe it would be clearer to say (rather than "Verify ..."), "The following steps verify ...".

doc/developer.texinfo
```diff
@@ -2446,34 +2436,99 @@ GNU Texinfo Manual (@uref{http://www.gnu.org/software/texinfo/manual/texinfo/tex
...
+
+@example
+make check-repo
+@end example
+
+@noindent
+The result of the command above should be:
```
mernst: This doesn't say what to do in that case.  How about moving the text about "must be committed to the repository" to here, saying that if it doesn't say "nothing to commit" then you should commit the files.  This seems like a clearer place to write those instructions, since this is the place where the user will take action.

doc/developer.texinfo
```diff
@@ -2446,34 +2436,99 @@ GNU Texinfo Manual (@uref{http://www.gnu.org/software/texinfo/manual/texinfo/tex
...
-@c Do not run this step on
-@c @samp{buffalo}.  For some reason the system software is not up to date.  Both
-@c @samp{bicycle} and @samp{tricycle} have been verified to work properly.
+@c [Time: 1 minute]
+
+Next, build everything to be distributed and copy it to the temporary staging area
+@c not @uref{...} because we don't want robots to notice and crawl the link
+(http://plse.cs.washington.edu/staging-daikon/).
+
+@example
```
mernst: How about combining this with the previous `@example` so that users can paste the whole thing at once rather than having to paste, then wait for completion, then paste again?

Makefile
markro49: added comments to Makefile; already noted in developer.texinfo.

Makefile
markro49: we just deleted staging. if no install and do cp it gets created in staging not staging/history.  if give staging/history as target of cp gets error does not exist. so i think we need the install

Makefile
markro49: done

Makefile
markro49: done

Makefile
markro49: parens are needed for the shell, but not a make file.  In fact, the cwd is reset at the start of each line of a recipe.

doc/developer.texinfo
markro49: done

doc/developer.texinfo
markro49: done

doc/developer.texinfo
markro49: done

Makefile
mernst: OK, I see.  Would "mkdir $STAGING_DIR" be enough?  That is, does this need to create the history/ subdirectory?

Also, maybe it's just me but I find using "install" counterintuitive when just creating a directory.  (That's what is happening here, right?)  I would use "mkdir" for that, because when I see "install" I assume it is copying a directory.

Makefile
mernst: I didn't know that.  Thanks for the info.

Makefile
markro49: i prefer mkdir too - just copying what was there.  i will change it to mkdir $(STAGING_DIR)

mernst: Thanks.  I think mkdir will be clearer.

PR #49, opened 2016-01-26
Type check under the new Lock Checker that provides value protection.

jthaine:
Note that the synchronization blocks I added are really questions ("is this the behavior that was intended?").


java/daikon/DynamicConstants.java
```diff
@@ -189,19 +190,19 @@ public Constant (VarInfo vi) {
 
 
     /*@EnsuresNonNullIf(result=true, expression="#1")*/
-    /*@Pure*/ public boolean equals (/*@Nullable*/ Object obj) {
+    /*@Pure*/ public boolean equals (/*>>>@GuardSatisfied Constant this,*/ /*>>>@GuardSatisfied @Nullable*/ Object obj) {
```
mernst: Use /*...*/ style comments whenever possible.  In particular, use two of them rather than one voodoo-style comment /*>>>...*/.

java/daikon/chicory/Runtime.java
```diff
@@ -129,7 +130,9 @@ public CallInfo (int nonce, boolean captured) {
     }
 
     /** Stack of active methods. **/
-    private static Map<Thread,Stack<CallInfo>> thread_to_callstack
+    // /*@GuardedBy("Runtime.class")*/ Map<Thread, /*@GuardedBy("Runtime.class")*/ Stack</*@GuardedBy("Runtime.class")*/ CallInfo>> is the ideal annotated type,
+    // however this is not possible since Map is not a proper parameterized type.
```
mernst: What is a "proper parameterized type"?

java/daikon/chicory/Runtime.java
```diff
@@ -146,43 +149,6 @@ private Runtime()
         throw new Error("Do not create instances of Runtime");
     }
 
-    /** Printf to dtrace file. **/
```
mernst: What is the reason for deleting these methods?
I guess it's because they are dead code?

java/daikon/dcomp/DCInstrument.java
```diff
@@ -1048,7 +1049,9 @@ public JavaClass instrument() {
     // Chicory runtime for this information.
     if (track_class) {
       // System.out.printf ("adding class %s to all class list%n", class_info);
-      daikon.chicory.Runtime.all_classes.add (class_info);
+      synchronized(daikon.chicory.Runtime.all_classes) {
+          daikon.chicory.Runtime.all_classes.add (class_info);
```
mernst: Inconsistent indentation

java/daikon/dcomp/DCInstrument.java
```diff
@@ -1207,7 +1210,9 @@ public JavaClass instrument_refs_only() {
     // Chicory runtime for this information.
     if (track_class) {
       // System.out.printf ("adding class %s to all class list%n", class_info);
-      daikon.chicory.Runtime.all_classes.add (class_info);
+        synchronized(daikon.chicory.Runtime.all_classes) {
```
mernst: Inconsistent indentation.

mernst:
Modulo a couple of remaining comments, this looks good.  Thanks!

java/daikon/PptTopLevel.java
```diff
@@ -2680,6 +2681,7 @@ public void addConditions(Splitter[] splits) {
   }
 
   /*@Pure*/
+  // TODO: Why is this method synchronized? There is no synchronization happening when proverStack is initialized.
```
mernst: I agree it's unnecessary.  Please remove the synchronization, and note it as an anecdote to report about finding a bug due to the process of annotating with the Lock Checker.

java/daikon/chicory/Runtime.java
mernst: What is a proper parameterized type, and why isn't Map one of them?

java/daikon/dcomp/DCInstrument.java
```diff
@@ -1048,7 +1049,9 @@ public JavaClass instrument() {
     // Chicory runtime for this information.
     if (track_class) {
       // System.out.printf ("adding class %s to all class list%n", class_info);
-      daikon.chicory.Runtime.all_classes.add (class_info);
+      synchronized(daikon.chicory.Runtime.all_classes) {
```
mernst: Here and below, add a space after `synchronized` so it doesn't look like a function call.

java/daikon/chicory/Runtime.java
jthaine: Yes - they are dead code, and I wanted to make sure they weren't type checked.

java/daikon/chicory/Runtime.java
jthaine: Sorry, this comment was misleading and I removed it. The issue is that the declaration of `Map` is `public interface Map<K extends Object, V extends Object>`.
Ideally we would be able to annotate it as
`public interface Map<K extends Object, V extends @GuardedByUnknown Object>` in order to allow Maps to store guarded values.
However, this would imply that Maps can only store and retrieve references to values but never dereference them, which is too restrictive.

java/daikon/chicory/ChicoryPremain.java
```diff
@@ -44,6 +44,18 @@
    **/
   private static boolean doPurity = false;
 
+  // This @SuppressWarnings is for the two lines
```
mernst: I suggest abstracting the two lines into their own method so that the scope of the @SuppressWarnings can be smaller.

java/daikon/chicory/ChicoryPremain.java
```diff
@@ -44,6 +44,18 @@
...
+  // Runtime.decl_writer = new DeclWriter (Runtime.dtrace);
+  // Runtime.dtrace_writer = new DTraceWriter (Runtime.dtrace);
+  //
+  // Runtime.dtrace is @GuardedBy("itself") because in the Runtime class,
+  // dtrace.close() is called only when the monitor of dtrace is held
```
mernst: The comment suggests that only the close() method is relevant.  But other operations matter too.

java/daikon/chicory/Runtime.java
```diff
@@ -439,7 +446,12 @@ public static void noMoreOutput() {
     // The incrementRecords method (which calls this) is called inside a
     // synchronized block, but re-synchronize just to be sure, or in case
     // this is called from elsewhere.
-    synchronized (Runtime.dtrace) {
+
+    // Ensure that the following method calls on dtrace all occur on the same
```
mernst: Here and two other places with a similar comment, the purpose if this code is not to ensure this property; we already believe the property to be true.  The code is a workaround because typechecker doesn't know that the side-effecting methods don't reset Runtime.dtrace.  If the field were reassigned, then the use of the local variable would mask the error rather than be useful.  Please update the comment.

java/daikon/chicory/Runtime.java
```diff
@@ -168,6 +173,7 @@ public TerminationMessage() {
   // the trace file, don't record it.
   private static boolean invokingPure = false;
 
+  /*@Holding("Runtime.class")*/
```
mernst: Why is the annotation on dontProcessPpts but not on startPure or endPure?  Those methods use the same field.

java/daikon/chicory/ChicoryPremain.java
jthaine: I clarified the comment - I meant the entire operation of closing a trace, which involves calling the println() and close() methods on it.

PR #57, opened 2016-05-31
Make Daikon type check after the fix to Checker Framework issue 657

jthaine:


mernst:
Many of these changes may now be obsolete.
We will try type-checking from scratch and create a new pull request with only the necessary changes.

PR #58, opened 2016-06-15
Change instances of the lock expression "itself" to "<self>"

jthaine:


mernst:
Looks fine, but typetools/checker-framework#799 must be merged before this is.

PR #63, opened 2016-07-26
Add Nullness Checker annotations

smillst:
Adds work arounds for  typetools/checker-framework#764 and typetools/checker-framework#862.  Also, adds `@RequiresNonNull` and `@EnsuresNonNull` annotations that should have appear before, but because of the issues mentioned above, the Nullness Checker did not issue errors.

### DIFF
--- a/java/daikon/FileIO.java
+++ b/java/daikon/FileIO.java
@@ -13,6 +13,7 @@ import daikon.config.Configuration;
 import daikon.derive.ValueAndModified;
 import daikon.diff.InvMap;
 import daikon.inv.Invariant;
+import daikon.suppress.NIS;
 import java.io.*;
 import java.net.*;
 import java.text.*;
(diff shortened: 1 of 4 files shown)


smillst:
Once this is merged, typetools/checker-framework#754 can be merged.

java/daikon/FileIO.java
```diff
@@ -1805,6 +1811,10 @@ public static void process_sample(
       return;
     }
 
+    // Workaround for Checker Framework issue #862
+    // https://github.com/typetools/checker-framework/issues/862
+    // Use NIS so that it is in scope to check the precondition below.
+    Object o = NIS.suppressor_map;
```
mernst: If this isn't used, I would name it something like dummy rather than o which is often used for real variables.

java/daikon/suppress/NIS.java
```diff
@@ -874,7 +875,9 @@ public static void remove_suppressed_invs(PptTopLevel ppt) {
   /**
    * Returns true if the specified class is an antecedent in any NI suppression
    */
-  /*@RequiresNonNull("suppressor_map")*/
+  //Use class name in front of static field to work around Checker Framework issue 764
```
mernst: Nit: separate text from comment marker.

mernst: Also, this comment makes good sense in the context of the diff but may be confusing to programmers reading the source code. I would omit it and mention the issue in the commit message rather than in a code  comment.

mernst:
Other than a couple of minor comments, this looks good and cam be merged. Thanks!

Issue #43, opened 2015-07-15
Rename command-line arguments for DynComp for C


mernst:
DynComp for C's command-line argument names are inconsistently named.
I propose the following renamings.

--gc-num-tags => --dyncomp-gc-num-tags

--no-dyncomp-gc => --dyncomp-gc-num-tags=0

--dyncomp-fast-mode => --dyncomp-approximate-literals

--separate-entry-exit-comp => --dyncomp-separate-entry-exit

--dyncomp-dataflow-only => --dyncomp-interactions=none
--dyncomp-dataflow-comp => --dyncomp-interactions=comparisons
--dyncomp-units => --dyncomp-dataflow=units
[does not exist; is the default] => --dyncomp-dataflow=all

markro49:
https://github.com/codespecs/fjalar/pull/17

completes this item

PR #65, opened 2016-08-09
Some changes to DynComp for C command line arguments.

markro49:
daikon issue #43

I made a couple of changes from the proposal.  Main one was to include 'units' with the other dyncomp-interactions types.


mernst:
Looks good to me.  Thanks!

PR #107, opened 2017-10-15
make goal name identification smarter

markro49:


tests/kvasir-tests/Makefile.common
```diff
@@ -75,11 +78,24 @@ endif
 ifeq ($(OS_REL),26)
 	GOAL := goal-Fedora26
 endif
+# other versions of Fedora will use goal-Fedora if it exists
+# otherwise the default goal file - which may be incorrect?
 endif
 
+# Function to compute goal filename extension.
```
mernst: Can you add at least a one-sentence comment saying what this is supposed to be doing?  That will help people reading it (or enable them to avoid reading the implementation).  Thanks.

tests/kvasir-tests/Makefile.common
```diff
@@ -56,13 +57,15 @@ ifeq (washington,$(findstring washington,$(shell uname -n)))
...
-	GOAL := goal-CentOS
+       GOAL := goal-CentOS-Std
 endif
 endif
 ifeq ($(OS_REL),7.3.1611)
 # Standard version of CentOS
-	GOAL := goal-CentOS
+       GOAL := goal-CentOS-Std
 endif
+# other versions of CentOS will use goal-CentOS if it exists
```
mernst: Nit:

tests/kvasir-tests/Makefile.common
```diff
@@ -56,13 +57,15 @@ ifeq (washington,$(findstring washington,$(shell uname -n)))
 	GOAL := goal-CentOS-UWCSE
 else
 # Standard version of CentOS
-	GOAL := goal-CentOS
+       GOAL := goal-CentOS-Std
```
mernst: I don't see any `.goal-CentOS-Std` files in the repository.  Why can't this remain just `.goal-CentOS`?

tests/kvasir-tests/Makefile.common
```diff
@@ -56,13 +57,15 @@ ifeq (washington,$(findstring washington,$(shell uname -n)))
...
 endif
 endif
 ifeq ($(OS_REL),7.3.1611)
 # Standard version of CentOS
-	GOAL := goal-CentOS
+       GOAL := goal-CentOS-Std
 endif
+# other versions of CentOS will use goal-CentOS if it exists
+# otherwise the default goal file - which may be incorrect?
```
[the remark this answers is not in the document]
markro49: done

tests/kvasir-tests/Makefile.common
markro49: There are a couple as part of this change.  They represent goal files that are unique to the standard version of CentOS (as we use on Travis).  With that change, we can use .goal-CentOS for goal files common to both the Travis and UWCSE versions of CentOS - of which there are several.  Similar to Fedora25 vs. Fedora.

tests/kvasir-tests/Makefile.common
markro49: done

tests/kvasir-tests/Makefile.common
mernst: I find the mention of a "standard" version of CentOS confusing.  Every version of CentOS  has a version number.  Shouldn't this be just `goal-CentOS` to act as the default?  Both a `.goal-CentOS` and a `.goal-CentOS-UWCSE` can exist, and I think that's clearer than the name "Std" that would require a definition.

markro49: The intent is to have a 'common' CentOS goal file for when CentOS and CentOS-UWCSE are identical.  (Like Fedora for Fedora24==Fedora25.)  Some name is then needed for when goal-CentOS-UWCSE==goal, but the non-UW version of CentOS is different.  So there are two choices: go back to having duplicate files or choose a different name for CentOS-Std.

mernst: > Some name is then needed for when goal-CentOS-UWCSE==goal, but the non-UW version of CentOS is different.

Thanks for the explanation.  This makes sense now.
I would call "the non-UW version of Centos" something more like CentOS-nonUWCSE rather than Centos-Std.
Thanks!

PR #127, opened 2018-03-20
remove plume.jar

markro49:
Remove last few uses of plume.jar and all references to it.  Makefiles still reference daikon/utils/plume-lib.


mernst:
Don't merge yet; perhaps reinstate uses of SimpleLog first, to reduce the size of the diffs.

PR #137, opened 2018-08-01
Update java lib jar files

markro49:
Several source code var/method name changes due to plume-util refactoring.
It looks to me that plume/options is okay being compiled with java 8 as it's build.gradle has the 1.7 compatibility statements.  I verified that all the .objects are V7.


markro49:
Forgot to mention:  The comments in java/lib/README about how to build java-getopt.jar.  I did not remove the .java files and things seem to be working ok.  If we go with this we will need to update the README.

mernst:
This looks great, thanks.
Please do simplify `java/lib/README`.
Then we should merge this before I merge my pull requests (because the two conflict, so you should go first; I know how to resolve the problems).

(I also changed some vague English instructions into copy-pasteable commands, which will be easier to use it the future.)

PR #157, opened 2018-12-13
Chicory variables flags bug fix

mmenarini:
While using the tool in our research project we found a bug in Chicory. 

There is a static field in the class DaikonVariableInfo called  empty_var_flags that is returned by the method get_var_flags().

All implementations of the DaikonVariableInfo override this method and should add the correct flags for the variable to the empty flag set. The bug is in FieldInfo. The code does not call clone() on the value returned by super.get_var_flags() and, therefore, the first time it is called it modifies the static set that is used as basis for all subsequent calls. 

The result is that when Chicory is creating declarations for multiple program points it ends up adding unnecessary flags.

Simplest solution is to add .clone() to FieldInfo.java line 197.  A better solution would be to move the .clone() form all the children to the abstract DaikonVariableInfo class.


mernst:
This change does not pass tests.  (See the continuous integration failure.)  Could you please fix that?  I'm not sure whether your changes introduce a bug or fix a bug (if the latter, then the tests should be changed).

Also, could you please contribute a test case that fails before the change but passes after the change (or at least that works differently before and after the change)?  Without that, it is hard to verify that this change fixes a real problem, and it is impossible to ensure that the changes are not reverted in the future.

Thanks!

mernst:
Thanks for the bug fix!  I appreciate it.

Sorry for the delay in merging -- I was waiting for a response and then this fell off my radar screen.  I fixed the tests and then merged.

PR #158, opened 2019-03-08
UWCSE upgrade from CentOS 7.5.1804 to 7.6.1810 needs baseline changes.

markro49:
New version of test results much closer to Fedora and Ubuntu.  I was able to delete several special case goal files.


markro49:
Update goal files for latest Fedora(28) and Ubuntu (1804) as well.

PR #163, opened 2019-06-07
Add rsync to ubuntu-jdk8 Dockerfile

VHellendoorn:
As described; patch is necessary to build Daikon Jar (see [Makefile, L602](https://github.com/codespecs/daikon/blob/master/Makefile#L602)). Please note that this same fix may need to be applied to some/all other Dockerfiles; happy to tag those commits on to this PR if found applicable.


mernst:
Thanks for the pull request.  Can you show why this is needed, such as how to reproduce a failure?
I ask because the Travis jobs that use a Docker image created from this Dockerfile succeed and use rsync:
https://travis-ci.org/codespecs/daikon/jobs/542019958#L925
https://travis-ci.org/codespecs/daikon/jobs/542019958#L927
https://travis-ci.org/codespecs/daikon/jobs/542019958#L928

mernst:
Per discussion on #164, it seems that perhaps you are using the Dockerfile to create a new Docker image rather than using the existing images that are available from Docker Hub.
So the bug is that this Dockerfile has become out of date.  At the last time it was used to create a Docker image, it worked correctly, but now it does not because the meaning of some packages has changed.
Is that correct?

VHellendoorn:
That sounds about right; my goal was to use these Dockerfiles to `docker build` an image within which Daikon can seamlessly be (re)built so I can quickly edit the code where needed (I actually just mount Daikon into the container). Rebuilding it myself seems like a reasonable use-case because it allows me to build straight from HEAD. Within that context, the dependency on `default-jdk` should be updated to point strictly at `jdk-8`.

mernst:
I agree.  Could you make that change?  And see whether anything besides rsync is necessary?
Thanks for figuring this out!

VHellendoorn:
So sorry to leave this stale; I was traveling. Unfortunately, I'm not sure how to resolve this: I hoped that simply swapping out `default-jdk` with e.g openjdk 8 would solve it, but none of the versions I have tried were able to build the project at all. I'm not sure which Java was originally successful (perhaps a specific Oracle version?). Tentatively I think we may have to shelve this and just keep the rsync observation in mind. Hopefully someone else can make it work with that info (who knows what kind of silly mistake I am making!)

Issue #162, opened 2019-06-07
Minor issues building Daikon with Dockerfile


VHellendoorn:
Using one of the provided Docker image (specifically, `Dockerfile-ubuntu-jdk8`) to recompile Daikon (in my case, rebuild the Daikon Jar), I ran into two small issues that may not necessarily be bugs (hence no PR) but may be worth inspecting:

- Line 57 of java/Makefile sets the `-Werror` flag. This causes the program to exit in case of compile warnings; since there are about 47 such warnings (many for seemingly insignificant issues), I had to remove this flag to allow full compilation.
- The build script relies on the rsync package, which is not included in the Dockerfile.
I can send a PR fixing either/both of the above, but figured I'd first check if/to what extent these are design decisions

mernst:
Thanks for pointing out these problems.  (And, I apologize that you ran into trouble.)

We would like to fix these things, so we would welcome pull requests (separately for the separate issues, please).

It's probably best to fix the warnings rather than removing `-werror`, if possible.  Can you tell me what they are?  I'm curious why the Travis jobs are not failing, since they do build `daikon.jar`.  Are you working from a release or from head?

The dockerfile should definitely include the rsync package.

Thanks!

VHellendoorn:
I'm building from HEAD (8bb826a); submitted PR #163 just now for the `rsync` issue.

As for `-Werror`, the warnings are all related to deprecation. Two samples demonstrating the two 
(very similar) categories:
```
./daikon/chicory/PureMethodInfo.java:70: warning: [deprecation] isAccessible() in AccessibleObject has been deprecated
    if (!meth.isAccessible()) {
             ^
./daikon/config/InvariantDoclet.java:54: warning: [removal] RootDoc in com.sun.javadoc has been deprecated and marked for removal
  public static boolean start(RootDoc doc) throws IOException {
```
I suppose the choice would be between suppressing `[deprecation]` and `[removal]` warnings or `-Werror`; I agree that keeping the latter is preferable. Please see PR #164

mernst:
I'm closing this because #163 and #164 have been closed.  Thanks for those pull requests!

PR #164, opened 2019-06-07
Disable 'deprecation' and 'removal' warnings to allow build to pass

VHellendoorn:
As described in codespecs/daikon#162: needed to allow a full build to pass because `-Werror` is set

### DIFF
--- a/java/Makefile
+++ b/java/Makefile
@@ -54,7 +54,7 @@ endif
 
 # Add flags like "-source 8" or "-target 8" to JAVAC_TARGET_FLAGS.
 # That variable appears last and overrides the values we provide here.
-JAVAC_ARGS ?= -g -XDignore.symbol.file -Xmaxerrs 100000 -Xmaxwarns 100000 -source 8 -target 8 -Werror -Xlint:-path -Xlint:-options -Xlint:-classfile ${JAVAC_TARGET_FLAGS}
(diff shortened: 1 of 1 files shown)


mernst:
The [Travis jobs](https://travis-ci.org/codespecs/daikon/builds/542889520) are all failing because they are run on JDK 8, but `-Xlint:removal` was added in JDK 9.  This also explains why we have not encountered the deprecation warnings:  at least the one in your example email was deprecated only in JDK 9, but we are testing on JDK 8.

This suggests that the pull request should check the version of Java, and add the two flags only for version 9+ (or check and not add them if the version is exactly 8).

Daikon does not fully support Java 9 yet.  The reason is that Java 9+ change the way that Java agents (run-time classfile instrumentation) work, and we have not yet adapted all tools shipped with Daikon.  (I expect Daikon proper to run on Java 9, but I haven't tested it myself.)

VHellendoorn:
That's very interesting; arguably this is an error in the corresponding Dockerfile (`ubuntu-jdk8`): it does not actually specify the Java version as JDK-8. Hence my confusion. I'm tentatively closing this, but please comment on #163 whether I should update said Dockerfile to explicitly pull `default-jdk-8`

mernst:
I agree with your assessment:  the Dockerfile has fallen out of date and needs to be updated.  Thanks for noticing and reporting this.

PR #174, opened 2019-07-05
(Minor fix) Fix verbose-mode printing in InvariantChecker

VHellendoorn:
In order to check an invariant against a trace file, InvariantChecker.java modifies the original invariant object (L515/518). When run in verbose mode, this produces incorrect debugging output; the invariant will be printed as having the set of values after adding the violating trace value. This PR corrects that by storing the unmodified string representation (only if in verbose mode) and printing that.


mernst:
Thanks for this improvement!

PR #214, opened 2020-03-09
Daikon support for Java 11

markro49:


java/daikon/chicory/ChicoryPremain.java
```diff
@@ -326,13 +326,17 @@ public static boolean isMethodPure(Member member) {
...
+   *   <li>the PLSE 6.4.1.1 release version (includes JDK 11 support)
    * </ul>
    *
-   * Currently, only DynComp cares about the LocalVariableGen fix, so we are looking for the offical
-   * 6.1 release version (or newer) and DynComp will check for the fixed version later. There are
-   * two classes present in 6.1 and subsequent releases that are not in previous versions. Thus, we
-   * can identify the correct version of BCEL by the presence of the class:
-   * org.apache.bcel.classfile.ConstantModule.class
+   * Currently, only DynComp cares about the LocalVariableGen fix or the JDK 11 support, so we are
+   * looking for the offical 6.1 release version (or newer) and DynComp will check for the 6.4.1.1
```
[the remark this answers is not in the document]
markro49: I have rewritten this comment to try to make it clearer what is going on:

   * Note that both Chicory and DynComp use the ChicoryLoader to load BCEL and to verify that the
   * version loaded is acceptable.  However, the official 6.1 release version is sufficient for
   * Chicory while DynComp requires the latest PLSE 6.4.1.1 version. Hence, this loader only
   * checks for the official 6.1 release version (or newer).  After loading BCEL, DynComp will
   * make an additional check to verify that the 6.4.1.1 version has been loaded.
   *
   * There are two classes present in 6.1 and subsequent releases that are not in
   * previous versions. Thus, we can identify the 6.1 version of BCEL by the presence of the
   * class: org.apache.bcel.classfile.ConstantModule.class.

java/daikon/chicory/ChicoryPremain.java
```diff
@@ -326,13 +326,17 @@ public static boolean isMethodPure(Member member) {
...
+   * version later. There are two classes present in 6.1 and subsequent releases that are not in
```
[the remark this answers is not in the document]
markro49: (see my reply to previous comment)

java/daikon/chicory/ChicoryPremain.java
```diff
@@ -326,13 +326,17 @@ public static boolean isMethodPure(Member member) {
...
+   * previous versions. Thus, we can identify the correct version of BCEL by the presence of the
```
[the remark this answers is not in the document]
markro49: (see my  reply to first comment)

java/daikon/dcomp/Premain.java
```diff
@@ -41,11 +45,55 @@
...
+
+  /** Set of classes known to cause problems when instrumented. */
+  protected static Set<String> problem_classes =
+      new HashSet<>(
+          Arrays.asList(
+              // Interfaces marked @FunctionalInterface should not be instrumented
+              // as they are used by LambdaExpressions.  Since BCEL won't let us
+              // inspect the annotations, we must manually add to problem classes.
+              "java.util.regex.Pattern$CharPredicate",
+              // <clinit> gets JNI during initialization.
```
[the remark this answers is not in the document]
markro49: Java Native Interface, but like most failures during JRE initialization, the error message is not very helpful:

Error: A JNI error has occurred, please check your installation and try again

java/Makefile
```diff
@@ -508,17 +507,34 @@ DCOMP_RT	:= dcomp-rt
...
+	  # Java 8 \
+	  $(JAR) cmf $(DCOMP_RT)/META-INF/MANIFEST.MF dcomp_rt.jar -C $(DCOMP_RT) .; \
+	else \
+	  # Java 9+ \
+	  # There appears to be a bug in the java reflection invoke code that it \
+	  # does not check that the arg list matches.  Hence, it tries to invoke \
+	  # our instrumented version from non instrumented code.  This causes a \
+	  # failure during java initialization of boot layer.  Thus we must remove \
+	  # the problem classes from dcomp_rt.jar. \
+	  rm $(DCOMP_RT)/java/lang/invoke/VarHandleInts?Field*.class; \
```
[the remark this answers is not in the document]
markro49: I could do that, but unfortunately, we still need the logic as there is no MANIFEST for 11 but there is for 8.  (Note the difference in the $(JAR) command line.)  I would suggest we leave as is and we can simplify when we drop Java 8.

java/daikon/chicory/Instrument.java
```diff
@@ -458,9 +463,7 @@ private ClassInfo instrument_all_methods(ClassGen cg, String fullClassName, Clas
             debug_instrument.log("arg_names(%d): %s%n", arg_names.length, names);
             debug_instrument.log("localvars(%d): %s%n", local_vars.length, locals);
             debug_instrument.log("Original code: %s%n", mg.getMethod().getCode());
-            debug_instrument.log("ClassInfo: %s%n", class_info);
-            debug_instrument.log("MethodGen: %s%n", mg);
-            dump_code_attributes(mg);
+            debug_instrument.log("%n");
```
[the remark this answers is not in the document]
markro49: just removing unnecessary/duplicate debugging output.

java/daikon/dcomp/Premain.java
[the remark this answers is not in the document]
mernst: Thanks.  "gets JNI" confused me.  "gets JNI error" or "gets a JNI error" would be more clear.

java/daikon/dcomp/Instrument.java
```diff
@@ -41,33 +41,73 @@ public Instrument() {
...
+            return null;
+          }
+        }
+
+        if (Premain.problem_classes.contains(className.replace('/', '.'))) {
+          if (DynComp.verbose) System.out.printf("Skipping problem class %s%n", className);
+          return null;
+        }
+
+        if (className.startsWith("java/lang/invoke/VarHandleInts")) {
```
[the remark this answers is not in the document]
markro49: That was on my todo list, but I got lazy because of the 'startsWith' - it covers 5 classes, I could add all of them to problem_classes as you suggest.  Would you like me to make the change?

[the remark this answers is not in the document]
mernst: Yes, please.  I think that will be clearer, and the documentation can be in a more discoverable place.  Thanks!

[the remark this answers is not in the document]
markro49: After some review and testing, we don't need this code in either file.  The problem only occurs at JRE startup and the fix in java/Makefile is designed for that case.

java/daikon/dcomp/Premain.java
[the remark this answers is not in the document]
markro49: done

tests/Makefile.common
```diff
@@ -219,35 +219,41 @@ default: help
 # is 1.n.n_n. For Java 9 or later it is n.n.n.n.
 # Currently, we only look for the release number, which is
 # the first 'n' in either case.
+# Note that version 11 (or 21 or ...) requires the 'start' anchor below
```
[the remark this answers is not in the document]
markro49: good idea - done

tests/Makefile.common
```diff
@@ -219,35 +219,41 @@ default: help
...
+startJVS := $(shell echo start$(JAVA_VERSION_STRING))
+ifneq (,$(findstring start1.,$(startJVS)))
   JAVA_RELEASE_NUMBER := $(shell echo $(JAVA_VERSION_STRING) | sed 's/1\.\(.\).*/\1/')
 else
   JAVA_RELEASE_NUMBER := $(shell echo $(JAVA_VERSION_STRING) | sed 's/\([0-9]*\)\..*/\1/')
 endif
 JAVA8_OR_NEWER := $(shell echo $(JAVA_RELEASE_NUMBER)\>7 | bc)
+JAVA11_OR_NEWER := $(shell echo $(JAVA_RELEASE_NUMBER)\>10 | bc)
 
 ifneq ($(JAVA8_OR_NEWER), 1)
```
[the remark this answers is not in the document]
markro49: done

tests/Makefile.common
```diff
@@ -219,35 +219,41 @@ default: help
...
 
 JAVA_VERSION_STRING := $(shell javac -version 2>&1 | head -1 | cut "-d " -f2)
-ifneq (,$(findstring 1.,$(JAVA_VERSION_STRING)))
+startJVS := $(shell echo start$(JAVA_VERSION_STRING))
+ifneq (,$(findstring start1.,$(startJVS)))
   JAVA_RELEASE_NUMBER := $(shell echo $(JAVA_VERSION_STRING) | sed 's/1\.\(.\).*/\1/')
 else
   JAVA_RELEASE_NUMBER := $(shell echo $(JAVA_VERSION_STRING) | sed 's/\([0-9]*\)\..*/\1/')
 endif
 JAVA8_OR_NEWER := $(shell echo $(JAVA_RELEASE_NUMBER)\>7 | bc)
```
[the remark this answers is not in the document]
markro49: done

tests/Makefile.common
```diff
@@ -219,35 +219,41 @@ default: help
...
 #
 # This goal_name does not apply to kvasir-tests which
 # have their own goal_name calculation.
 
-goal_name = $(if $(findstring 8, $(JAVA_RELEASE_NUMBER)), \
+goal_name = $(if $(findstring 0, $(JAVA11_OR_NEWER)), \
 				$(1).goal, \
-				$(if $(shell if [ -f $(1).goal.j9 ] ; then echo true; fi), \
-					$(1).goal.j9, \
+				$(if $(shell if [ -f $(1).goal.j11 ] ; then echo true; fi), \
```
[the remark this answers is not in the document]
markro49: I agree in principle, but a little messy and would prefer to wait until this pull request is complete.

java/daikon/dcomp/BuildJDK.java
```diff
@@ -275,103 +222,219 @@ void copyStreams(InputStream fis, OutputStream fos) throws Exception {
...
-            || (skip_object && entryName.equals("java/lang/Object.class"))
-            || (!prefix.equals("") && !entryName.startsWith(prefix))) {
-          File destfile = new File(entryName);
-          if (destfile.getParent() == null) {
-            System.out.printf("Skipping file %s%n", destfile);
+
+        if (className.endsWith(".class")) all_classes.add(className.replace(".class", ""));
+        if (!className.endsWith(".class") || className.equals("java/lang/Object.class")) {
+          File destfile = new File(className);
+          if (destfile.getParent() == null || BcelUtil.javaVersion > 8) {
```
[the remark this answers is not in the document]
markro49: added comments

java/daikon/dcomp/BuildJDK.java
```diff
@@ -90,109 +59,72 @@ protected boolean exists(String libRt) {
...
-   *   <dd>jarfile to process
-   *   <dt>dest
-   *   <dd>destination directory in which to place instrumented classes
-   *   <dt>prefix
-   *   <dd>optional prefix of classes to be translated
-   * </dl>
-   *
-   * Instruments each class file in jarfile that begins with prefix and puts the results in dest.
-   */
+  private Map<String, InputStream> classmap = new HashMap<>();
```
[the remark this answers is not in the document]
markro49: changed to class_stream_map and added explanation

java/daikon/dcomp/BuildJDK.java
```diff
@@ -33,55 +40,17 @@
  */
```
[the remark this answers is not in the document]
markro49: done

java/daikon/dcomp/BuildJDK.java
```diff
@@ -275,103 +222,219 @@ void copyStreams(InputStream fis, OutputStream fos) throws Exception {
...
+
+    FileSystem fs = FileSystems.getFileSystem(URI.create("jrt:/"));
+    Path modules = fs.getPath("/modules");
+    System.out.printf("using modules file %s%n", System.getProperty("java.home") + "/lib/modules");
+    try (DirectoryStream<Path> directoryStream = Files.newDirectoryStream(modules, "java.base*")) {
+      for (Path module : directoryStream) {
+        translate_modules_directory(module, module.toString().length());
+      }
+    } catch (IOException e) {
+      e.printStackTrace();
```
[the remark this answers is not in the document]
markro49: Added comment describing use of quit_if_error and that it was tested in DCInstrument.  Also changed e.printStackTrace to a throw.

java/daikon/dcomp/BuildJDK.java
```diff
@@ -275,103 +222,219 @@ void copyStreams(InputStream fis, OutputStream fos) throws Exception {
...
+
+  void translate_modules_directory(Path path, int modulePrefixLength) {
+
+    if (Files.isDirectory(path)) {
+      try (DirectoryStream<Path> directoryStream = Files.newDirectoryStream(path)) {
+        for (Path subpath : directoryStream) {
+          translate_modules_directory(subpath, modulePrefixLength);
+        }
+      } catch (IOException e) {
+        e.printStackTrace();
```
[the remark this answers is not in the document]
markro49: changed as noted above

java/daikon/dcomp/BuildJDK.java
```diff
@@ -275,103 +222,219 @@ void copyStreams(InputStream fis, OutputStream fos) throws Exception {
...
+            if (verbose) System.out.printf("Skipping file %s%n", destfile);
             continue;
           }
-          File dir = new File(dfile, destfile.getParent());
+          File dir = new File(dest_dir, destfile.getParent());
           dir.mkdirs();
           File destpath = new File(dir, destfile.getName());
           if (verbose) System.out.println("Copying Object or non-classfile: " + destpath);
-          copyStreams(jfile.getInputStream(entry), new FileOutputStream(destpath));
+          copyStreams(classmap.get(className), new FileOutputStream(destpath));
```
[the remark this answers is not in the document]
markro49: done

java/daikon/dcomp/BuildJDK.java
```diff
@@ -90,288 +57,380 @@ protected boolean exists(String libRt) {
...
       DCInstrument.restore_static_map(new File(dest_dir, static_map_fname));
-      System.out.printf("Restored %d entries in static map%n", DCInstrument.static_map.size());
+      System.out.printf("Restored %d entries in static map.%n", DCInstrument.static_map.size());
 
       // Read in each specified classfile
-      Map<String, JavaClass> classmap = new LinkedHashMap<>();
       for (File class_file : class_files) {
-        if (class_file.toString().endsWith("java/lang/Object.class")) {
-          System.out.printf("Skipping %s%n", class_file);
+        String className = class_file.toString();
```
mernst: This variable named `className` does not hold a class name.  It holds a file name.

java/daikon/dcomp/BuildJDK.java
```diff
@@ -90,288 +57,380 @@ protected boolean exists(String libRt) {
...
+      if (target_args.length < 2) {
         System.out.println("must specify classfiles to instrument");
         options.printUsage();
-        return false;
+        System.exit(1);
       }
     } else {
-      if (target_args.length < 2) {
+      if (target_args.length < 1) {
         System.out.println("must specify source jar and destination dir");
```
mernst: The two arguments seem to be in the wrong order:  doesn't the destination directory have to be first?

java/daikon/dcomp/BuildJDK.java
```diff
@@ -90,288 +57,380 @@ protected boolean exists(String libRt) {
...
-    fos.close();
   }
 
-  void translate_classes(JarFile jfile, String dest, String prefix, String prefixOfFilesToInclude)
-      throws java.io.IOException {
-
-    // Map from classname to class so we can find out information about
-    // classes we have not yet instrumented.
-    Map<String, JavaClass> classmap = new LinkedHashMap<>();
+  void translate_jar() throws java.io.IOException {
```
mernst: What does this method do?  It's not documented.  (I realize this isn't your fault.)

java/daikon/dcomp/BuildJDK.java
```diff
@@ -90,288 +57,380 @@ protected boolean exists(String libRt) {
...
+        // Get the InputStream for this file
+        InputStream is = jfile.getInputStream(entry);
+        class_stream_map.put(entryName, is);
+      }
+    } catch (Exception e) {
+      throw new Error("Problem while reading " + jar_name, e);
+    }
+  }
+
+  void translate_modules() throws java.io.IOException {
```
mernst: What does this method do?  It's not documented.  Every method should be documented.
The name is rather confusing because this doesn't seem to do any translation, it just collects information.

java/daikon/dcomp/BuildJDK.java
```diff
@@ -90,288 +57,380 @@ protected boolean exists(String libRt) {
...
+    try (DirectoryStream<Path> directoryStream = Files.newDirectoryStream(modules, "java.base*")) {
+      for (Path module : directoryStream) {
+        translate_modules_directory(module, module.toString().length());
+      }
+    } catch (IOException e) {
+      throw new Error(e);
+    }
+  }
+
+  void translate_modules_directory(Path path, int modulePrefixLength) {
```
mernst: Please document every method.  Does this actually do any translation?

mernst:
Could you please add some documentation to help me understand and review the code?

markro49:
I will start on adding to and improving the documentation tomorrow.

java/daikon/dcomp/BuildJDK.java
```diff
@@ -90,288 +57,380 @@ protected boolean exists(String libRt) {
...
   public static void main(String[] args) throws IOException {
 
-    System.out.println("Starting at " + new Date());
+    System.out.println("BuildJDK starting at " + new Date());
 
-    Options options = new Options(synopsis, BuildJDK.class, DynComp.class);
-    // options.ignore_options_after_arg (true);
+    Options options =
+        new Options(
+            "daikon.BuildJDK [options] dest [classfiles...]", BuildJDK.class, DynComp.class);
```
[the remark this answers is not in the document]
markro49: There is an Options comment for classfiles at line 50.  I will be adding several lines to the BuildJDK comments starting at line 34 to better describe the two modes of operation.

markro49:
This commit is to get the javadoc test passing.  It does not address all of the BuildJDK comments.

markro49:
don't understand - it compiled fine on my javac 1.8.0_211

mernst:
The problem is use of non-ASCII characters.

java/daikon/dcomp/BuildJDK.java
markro49: error message was not updated.  should be "must specify destination dir"
fixed

markro49:
DCInstrument still needs javadoc fixes, but your questions about BuildJDK have been addressed in this commit.

java/daikon/dcomp/BuildJDK.java
```diff
@@ -3,375 +3,500 @@
...
-        }
-      }
-      return jarFilename;
-    }
+  /**
+   * Instrument the class files given on the command line instead of the ones in the Java runtime
+   * library.
+   */
+  @Option("Instrument only the class files given on the command line")
+  public static boolean classfiles = false;
```
mernst: I don't think this command-line argument is necessary.  It can be inferred from the presence of command-line arguments.  In particular:  if there is one command-line argument, then instrument the JDK, and if there are more, instrument the given classes.  Little logic and documentation need to be changed other than removing this command-line argument.

java/daikon/dcomp/BuildJDK.java
```diff
@@ -3,375 +3,500 @@
...
   private static boolean verbose = false;
 
-  /** Whether or not to instrument java.lang.Object */
-  private static boolean skip_object = true;
-
+  /** Number of class files processed; used for progress display. */
   private int _numFilesProcessed = 0;
 
-  private static String static_map_fname = "dcomp_jdk_static_map";
+  /** Name of entry in output jar containing the statics map. */
```
mernst: I first thought fname stood for "field name", but I think it's actually short for "file name".  I would write it out.  I was also confused by the documentation saying it's an "entry".

java/daikon/dcomp/BuildJDK.java
```diff
@@ -3,375 +3,500 @@
...
+  private static String static_field_id_fname = "dcomp_jdk_static_field_id";
 
+  /** Names of all the classes processed. */
```
mernst: In what format?  I think these might not be class names, but actually be relative (as opposed to absolute) class file names, though with ".class" removed from the end.
(That's a somewhat funny string format; what is the reason for it?)

java/daikon/dcomp/BuildJDK.java
```diff
@@ -3,375 +3,500 @@
...
   private static List<String> all_classes = new ArrayList<>();
 
+  /** Names of all methods that DCInstrument could not process. Should be empty. */
```
mernst: Are these simple names, or do they include class names, and if so, what is the format?

java/daikon/dcomp/BuildJDK.java
```diff
@@ -3,375 +3,500 @@
...
-        "com.sun.tools.javac.jvm.ClassReader.readInputStream",
-        "com.sun.tools.javac.processing.ServiceProxy.parse",
-        "com.sun.tools.javac.zip.ZipFileIndex$DirectoryEntry.initEntries",
-        "com.sun.tools.javac.zip.ZipFileIndex.readIndex",
-        "com.sun.tools.javac.zip.ZipFileIndex.writeIndex",
-        */
-      };
+  /**
+   * A list of methods known to cause DCInstrument to fail. This is used to remove known problems
+   * from the list of failures displayed at the end of BuildJDK's execution.
```
mernst: In what format?  (I can see it in the diff, but with an empty list someone reading the code can't know it.)

java/daikon/dcomp/BuildJDK.java
```diff
@@ -3,375 +3,500 @@
...
-      System.out.printf("Found %d statics%n", DCInstrument.static_map.size());
-      DCInstrument.save_static_map(new File(dest_dir, static_map_fname));
+      System.out.printf("Found %d statics.%n", DCInstrument.static_field_id.size());
+      DCInstrument.save_static_field_id(new File(dest_dir, static_field_id_fname));
 
       // Write out the list of all classes in the jar file
       File jdk_classes_dir = new File(dest_dir, "java/lang");
       File jdk_classes_file = new File(jdk_classes_dir, "jdk_classes.txt");
-      PrintWriter pw = new PrintWriter(jdk_classes_file, UTF_8.name());
       System.out.printf("Writing all classes to %s%n", jdk_classes_file);
```
mernst: Should this be "class names" rather than "classes"?

java/daikon/dcomp/BuildJDK.java
```diff
@@ -3,375 +3,500 @@
...
 
-  void translate_classes(JarFile jfile, String dest, String prefix, String prefixOfFilesToInclude)
-      throws java.io.IOException {
-
-    // Map from classname to class so we can find out information about
-    // classes we have not yet instrumented.
-    Map<String, JavaClass> classmap = new LinkedHashMap<>();
+  /**
+   * For Java 8 the Java runtime is located in rt.jar. This method creates an InputStream for each
+   * class in rt.jar and saves this information in {@link #class_stream_map}.
```
mernst: I don't think there is any need for a `class_stream_map` field.
Instead, this method could create and return the map, and the client (that is, `main`) could pass the map to `instrument_classes`.  That would clarify exactly where and how the map is used, and eliminate an unneeded field.

java/daikon/dcomp/BuildJDK.java
```diff
@@ -3,375 +3,500 @@
...
+   * For Java 9+ the Java runtime is located in a series of modules. At this time, we are only
+   * pre-instrumenting the java.base module. This method initializes the DirectoryStream used to
+   * explore java.base. It calls gather_runtime_from_modules_directory to process the directory
+   * structure.
+   */
+  void gather_runtime_from_modules() {
+
+    FileSystem fs = FileSystems.getFileSystem(URI.create("jrt:/"));
+    Path modules = fs.getPath("/modules");
+    System.out.printf("using modules file %s%n", java_home + "/lib/modules");
```
mernst: I think this is a directory, not a (plain) file.

java/daikon/dcomp/BuildJDK.java
```diff
@@ -3,375 +3,500 @@
...
+    try (DirectoryStream<Path> directoryStream = Files.newDirectoryStream(modules, "java.base*")) {
+      for (Path module : directoryStream) {
```
mernst: Since I think this is a directory, I would name it `moduleDir` rather than `module`.

java/daikon/dcomp/BuildJDK.java
```diff
@@ -3,375 +3,500 @@
...
+  public static List<String> known_uninstrumentable_methods =
+      Arrays.asList(
+          // None at present
+          );
+
+  /**
+   * We want to share code to read and instrument the Java class file members of a jar file (JDK 8)
+   * or a module file (JDK 9+). However, jar files and module files are located in two completely
+   * different file systems. So we open an InputStream for each class file we wish to instrument and
+   * save it the the class_stream_map. From that point the code to instrument a class file can be
```
mernst: What is the format of keys in this map?  I think they might be file names instead of class names, but I'm not sure.

java/daikon/dcomp/BuildJDK.java
```diff
@@ -3,375 +3,500 @@
...
+  }
+
+  void instrument_classes(File dest_dir) {
+
+    try {
+      // Create the destination directory
+      dest_dir.mkdirs();
+
+      // Process each file.
+      for (String className : class_stream_map.keySet()) {
```
mernst: I don't think these are class names.  I think they are file names or jar file entry names.  Please adjust variable name and/or documentation accordingly.

java/daikon/dcomp/BuildJDK.java
```diff
@@ -3,375 +3,500 @@
...
               new String[0]);
+      dcomp_marker.setMinor(0);
+      dcomp_marker.setMajor(java_class_version);
       dcomp_marker
           .getJavaClass()
           .dump(
               new File(
-                  dest, "java" + File.separator + "lang" + File.separator + "DCompMarker.class"));
+                  dest_dir_name,
+                  "java" + File.separator + "lang" + File.separator + "DCompMarker.class"));
```
mernst: I suggest abstracting this out into a new method, since it is now used 5 times in this class.  That will shorten and simplify the code.

java/daikon/dcomp/BuildJDK.java
```diff
@@ -380,12 +505,13 @@ private void processClassFile(Map<String, JavaClass> classmap, File dfile, Strin
       inst_jc = dci.instrument_jdk();
     }
     skipped_methods.addAll(dci.get_skipped_methods());
-    File classfile = new File(classname.replace('.', '/') + ".class");
+    // File classfile = new File(classname.replace('.', '/') + ".class");
```
mernst: You can probably remove this comment.

java/daikon/dcomp/BuildJDK.java
markro49: Sounds like a good idea.  I skipped this one for now to get everything else checked in.

java/daikon/dcomp/BuildJDK.java
markro49: I went ahead and made these changes - and perhaps I did not do it the way you expected, but I found the results to be a bit awkward with several more lines of code and extra arguments.

java/daikon/dcomp/BuildJDK.java
markro49: Simplified code and looks good - thanks for suggestion.

java/daikon/dcomp/BuildJDK.java
```diff
@@ -3,375 +3,496 @@
...
+   * structure.
+   *
+   * @return a map from class file name to the associated InputStream
+   */
+  Map<String, InputStream> gather_runtime_from_modules() {
+
+    Map<String, InputStream> class_stream_map = new HashMap<>();
+    FileSystem fs = FileSystems.getFileSystem(URI.create("jrt:/"));
+    Path modules = fs.getPath("/modules");
+    System.out.printf("using modules directory %s%n", java_home + "/lib/modules");
```
mernst, line 284: The filename printed here differs from the URI used on the previous two lines.  Maybe that's OK, I'm just pointing it out.

java/daikon/dcomp/BuildJDK.java
mernst: Will you make this change before we merge?  I think it will make the changes smaller and easier to read.
(No problem with doing the other changes first.)

java/daikon/dcomp/DCInstrument.java
```diff
@@ -1935,8 +2076,12 @@ InstructionList xform_inst_refs_only(MethodGen mg, InstructionHandle ih, Operand
    * Adds a call to DCruntime.exit() at each return from the method. This call calculates
    * comparability on the daikon variables. It is only necessary if we are tracking comparability
    * for the variables of this method.
+   *
+   * @param mg method to modify
+   * @param mi MethodInfo for method
+   * @param method_info_index index for MethodInfo
    */
-  public void add_exit(MethodGen mg, MethodInfo mi, int method_info_index) {
+  void add_exit(MethodGen mg, MethodInfo mi, int method_info_index) {
```
mernst, line 2084: Many methods in DCInstrument were changed from `public` to package-protected.  (That was the reason for the need for a lot of Javadoc.)  I'm curious about the reason for this change.

markro49: The methods modified are all defined and used entirely within DCInstrument.   I couldn't see a use case for them to be public.

mernst: I see.  In that case, should they be private?

markro49: The coding style seemed to be use private for data but not methods.  There are a few exceptions, of course.  I'm happy to change to a standard/recommended style.

java/daikon/dcomp/BuildJDK.java
```diff
@@ -3,375 +3,496 @@
...
-    }
-
-    /** @return true iff libRt exists in the filesystem */
-    protected boolean exists(String libRt) {
-      return new File(libRt).exists();
-    }
-  }
+  /**
+   * Used when testing to terminate processing if an error occurs. This flag is tested in
+   * DCInstrument.
```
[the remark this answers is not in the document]
markro49: But I would need to add DCInstrument.class to the new Opions(...), correct?

java/daikon/dcomp/BuildJDK.java
```diff
@@ -3,375 +3,496 @@
...
-      File dest_dir = new File(cl_args[1]);
-      File[] class_files = new File[cl_args.length - 2];
-      for (int ii = 2; ii < cl_args.length; ii++) {
-        class_files[ii - 2] = new File(cl_args[ii]);
-      }
+    if (cl_args.length > 1) {
 
-      BuildJDK build = new BuildJDK();
+      // Arguments are <destdir> <classfiles>...
+      File[] class_files = new File[cl_args.length - 1];
```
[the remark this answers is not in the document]
markro49: There is a conversion from String to File in the loop - don't see how to do that with either of those copy techniques.

But wait, the first statement in the class_files loop is to convert back to String!
Oh man, that's sad; but as you said - not my code ...

java/daikon/dcomp/BuildJDK.java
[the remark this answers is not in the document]
mernst: Yes, that is correct.

java/daikon/dcomp/BuildJDK.java
```diff
@@ -3,375 +3,496 @@
...
+    File dest_dir = new File(cl_args[0]);
 
-      // Arguments are <srcdir> <destdir> <classfiles>...
-      File src_dir = new File(cl_args[0]);
-      File dest_dir = new File(cl_args[1]);
-      File[] class_files = new File[cl_args.length - 2];
-      for (int ii = 2; ii < cl_args.length; ii++) {
-        class_files[ii - 2] = new File(cl_args[ii]);
-      }
+    if (cl_args.length > 1) {
```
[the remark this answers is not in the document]
markro49: This is not too hard to do.  One thing I noticed is the chunk of code at the end of instrument_classes would have to be pulled out into a separate method (does not apply to classes on cmdline case), but that fits nicely with the cleanup we have planned for that code.

java/daikon/dcomp/BuildJDK.java
```diff
@@ -3,375 +3,496 @@
...
-      PrintWriter pw = new PrintWriter(jdk_classes_file, UTF_8.name());
-      System.out.printf("Writing all classes to %s%n", jdk_classes_file);
-      for (String classname : all_classes) {
-        pw.println(classname);
+      File jdk_classes_file = new File(dest_dir, "java/lang/jdk_classes.txt");
+      System.out.printf("Writing a list of class names to %s%n", jdk_classes_file);
+      // Class names are written in internal form.
+      try (PrintWriter pw = new PrintWriter(jdk_classes_file, UTF_8.name())) {
+        pw.println("no_primitives: " + DynComp.no_primitives);
+        for (String classname : all_classes) {
```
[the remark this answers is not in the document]
markro49: done - not hard at all

java/daikon/dcomp/BuildJDK.java
markro49: java_home+"/lib/modules" is the host file system equivalent of "jrt:/modules".  I will add a comment to that effect.  Does this need to be displayed for the user?  I'm thinking no.

mernst: Sounds good, thanks.

java/daikon/dcomp/DCInstrument.java
mernst: A general rule of thumb is to make all data private, to avoid data corruption by clients directly changing fields without maintaining invariants.

Methods should be made public/protected/private/package-private depending on their intended use; there isn't a general rule for Java (nor one for this codebase that I am aware of) that requires methods not to be private.

java/daikon/dcomp/BuildJDK.java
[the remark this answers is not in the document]
markro49: i moved the call to print_skipped_methods after the end of the if statement.

java/daikon/dcomp/BuildJDK.java
```diff
@@ -3,375 +3,461 @@
...
-            || (!prefix.equals("") && !entryName.startsWith(prefix))) {
-          File destfile = new File(entryName);
-          if (destfile.getParent() == null) {
-            System.out.printf("Skipping file %s%n", destfile);
+
+        if (!classFileName.endsWith(".class") || classFileName.equals("java/lang/Object.class")) {
+          File destfile = new File(classFileName);
+          // Copy non-.class files (and Object.class) unchanged (JDK 8).
+          // For JDK 9+ we do not copy as these items will be loaded from the original module file.
+          if (destfile.getParent() == null || BcelUtil.javaVersion > 8) {
```
mernst, line 345: When is `destfile.getParent()` null?  The comments don't say anything about that part of the test.

java/daikon/dcomp/BuildJDK.java
```diff
@@ -3,375 +3,461 @@
...
       }
+    } catch (Exception e) {
+      throw new Error(e);
+    }
+  }
 
-      // Create the DcompMarker class (used to identify instrumented calls)
-      ClassGen dcomp_marker =
+  /**
+   * Add abstract interface classes needed by DynComp runtime.
```
mernst, line 382: Can you clarify this comment?  What are they added to?  Should the method be named "create" rather than "add"?

mernst:
@markro I broke the build.  Sorry about that!  I am investigating.

mernst:
@markro It seems I did not break the build, but that job is nondeterministic.

I thought I had broken the build because re-running CI twice led to failures.  However, running a third time made it pass.  So, the kvasir_centos_jdk8 job is nondeterministic.

More evidence is the fact that there are two branches
daikon-branch-j11-support-bug-investigation
daikon-branch-j11-support-bug-investigation-duplicate
that have identical contents but different history (and therefore different commit IDs).
One of them passes CI but the other fails:
https://dev.azure.com/mernstdaikon/daikon/_build?definitionId=1&_a=summary&view=branches

Could you look into this nondeterminism?  That can happen after this pull request is merged, but we should do it.  Otherwise it will waste our time when the tests fail.

java/daikon/dcomp/BuildJDK.java
markro49: test is to skip any non ".class" file at root of jar entries.  In a normal rt.jar file there are none.  Don't know why Ben put this there.  Though it does protect the use in the 'new File" call just below the end of the if.

Comment or remove it?  I lean towards remove it.  Its inside a try-catch that will report error if not a well formed rt.jar.

java/daikon/dcomp/BuildJDK.java
markro49: changed comment to:
* Create abstract interface classes needed by DynComp runtime and write to destDir.

java/daikon/dcomp/BuildJDK.java
mernst: The `if (!classFileName.endsWith(".class")` test just above skips non-`.class` files.
This test skips **`.class`** files at the root, I think.
I made changes in fd0bc8e72; could you review that commit?

mernst:
This is passing CI and is ready to merge after @markro49 reviews my BuildJDK.java changes.

markro49:
Your changes to BuildJDK look good.  Let's merge.

PR #222, opened 2020-03-12
Refactorings for Java 11 support

mernst:


mernst:
@markro49 The point of assigning this pull request to you is to verify that this does not move code outside the try block that, for some subtle reason, needs to be within the try block.

markro49:
getClassLoader can throw a SecurityException.  I don't know if this is an issue or not.  It could be that if the user program under analysis installs a security manager then the DynComp instrumenter may have problems anyway,  I don't have enough knowledge in this area to answer the question.

mernst:
So, it sounds from your answer like we should close this pull request without merging the change, because there is an important side effect that must happen within the scope of the try block.
Is that right?

markro49:
Code as is might catch SecurityException and throw RuntimeException ("Unexpected error loading Instrument", e). (Where e is SecurityException.)  In same circumstances, the revised code would just throw the SecurityException.  I think I'm OK with that - your call.  I do like your code simplification.  It's just a question of inside or side the try/catch.

PR #224, opened 2020-03-12
Change return type of find_enabled

mernst:


mernst:
> There are classes that inherit a method and ideally would redeclare it to mark types as `@Nullable`.

I agree, but this is not currently possible.  See #3094.

PR #232, opened 2020-04-04
Adapt to typetools/checker-framework#3226

xingweitian:
This PR tries to merge https://github.com/mernst/daikon/tree/field-set-nullness into the master branch.


xingweitian:
@mernst Should we also suppress `nullness:unneeded.suppression`?

mernst:
> @mernst Should we also suppress `nullness:unneeded.suppression`?

No.  Doing so would cause Daikon to fail type-checking once https://github.com/typetools/checker-framework/pull/3226 is merged.  Daikon type-checking passes in that PR, and therefore it will pass once that PR and this one are merged.

The failure here just says that Daikon has been changed incompatibly with the current Checker Framework, which is expected.

PR #256, opened 2020-06-23
Rename java.daikon.dcomp.Test to  java.daikon.dcomp.DcompTest

mernst:


mernst:
Merging despite build failures because the only build failures are due to missing documentation in a test full of nonsense code.

PR #271, opened 2020-08-07
Add DynComp support for JUnit test framework.

markro49:
Have DynComp recognize a JUnit test class and process specially.  Also add a new test: daikon-test/junit.


markro49:
Will not fix JDK 11 junit.  Need to update Annotation code.

PR #272, opened 2020-08-09
Suppress instanceof annotation check warnings

aditya3434:
Adding a ```@SuppressWarnings("allcheckers:operand.instanceof.subtype")``` line in OneOf.java.jpp to just check the data type (and not the annotations) of the ```state``` variable in the ```boolean state_match()``` method. This was done to bypass the ```instanceof``` error and pass the daikon tests in the pull request #3481 in typetools/checker-framework (https://github.com/typetools/checker-framework/pull/3481).

### DIFF
--- a/java/daikon/inv/unary/OneOf.java.jpp
+++ b/java/daikon/inv/unary/OneOf.java.jpp
@@ -2118,9 +2118,9 @@ public String format_csharp_contract(@GuardSatisfied ONEOFSCALAR this) {
     if (num_elts == 0) {
       return false;
     }
-    @SuppressWarnings("allcheckers:operand.instanceof.subtype") //  Warnings are suppressed because only the data type of the state needs to be checked and not the annotations
-    boolean stateTypeMatch = !(state instanceof INT[]);
-    if (stateTypeMatch) {
+    @SuppressWarnings("allcheckers:instanceof.unsafe") //  Only the data type of the state needs to be checked and the annotations must be ignored, hence, the qualifier is irrelevant
+    boolean stateIsNotArray = !(state instanceof INT[]);
+    if (stateIsNotArray) {
       System.out.println("state is of class '" + state.getClass().getName()
                           + "'");
     }
--- a/java/daikon/inv/unary/OneOf.java.jpp
+++ b/java/daikon/inv/unary/OneOf.java.jpp
@@ -2118,7 +2118,6 @@ public String format_csharp_contract(@GuardSatisfied ONEOFSCALAR this) {
     if (num_elts == 0) {
       return false;
     }
-    @SuppressWarnings("allcheckers:instanceof.unsafe") //  Only the data type of the state needs to be checked and the annotations must be ignored, hence, the qualifier is irrelevant
     boolean stateIsNotArray = !(state instanceof INTARRAY);
     if (stateIsNotArray) {
       // Daikon is about to crash.  Produce some debugging output.
--- a/java/daikon/inv/unary/OneOf.java.jpp
+++ b/java/daikon/inv/unary/OneOf.java.jpp
@@ -2113,7 +2113,7 @@ public String format_csharp_contract(@GuardSatisfied ONEOFSCALAR this) {
    * example if x = 1 and the state contains 1 and 2, true will be returned.
    */
   @Override
-  @SuppressWarnings("interning:operand.instanceof.subtype")
+  @SuppressWarnings("allcheckers:operand.instanceof.subtype")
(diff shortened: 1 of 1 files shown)


mernst:
The pull request description does not make sense, since there is no issue #3033 in this repository.  Please describe the pull request in a way that makes sense to users of this repository.

java/daikon/inv/unary/OneOf.java.jpp
```diff
@@ -2113,6 +2113,7 @@ public String format_csharp_contract(@GuardSatisfied ONEOFSCALAR this) {
    * example if x = 1 and the state contains 1 and 2, true will be returned.
    */
   @Override
+  @SuppressWarnings("allcheckers:operand.instanceof.subtype")
```
mernst, line 2116: Every `@SuppressWarnings` annotation should be accompanied by a description of the warning and a justification of why the error is a false positive.

aditya3434: I have limited the scope of the ```@SuppressWarnings``` by introducing a boolean variable and added the justification of the suppression as a comment in my latest commit.

aditya3434:
> The pull request description does not make sense, since there is no issue #3033 in this repository. Please describe the pull request in a way that makes sense to users of this repository.

I have changed the PR description and title so that none of them reference issue#3033 which is absent in this repository.

java/daikon/inv/unary/OneOf.java.jpp
```diff
@@ -2118,8 +2118,9 @@ public String format_csharp_contract(@GuardSatisfied ONEOFSCALAR this) {
     if (num_elts == 0) {
       return false;
     }
-
-    if (!(state instanceof INT[])) {
+    @SuppressWarnings("allcheckers:operand.instanceof.subtype")
+    boolean stateTypeMatch = !(state instanceof INT[]); //  Warnings are suppressed because only the data type of the state needs to be checked and not the annotations
```
mernst, line 2122: Please place the comment on the line with `@SuppressWarnings`, because the comment is about the warning suppression.

aditya3434: I have changed the comment's location in my latest commit.

java/daikon/inv/unary/OneOf.java.jpp
```diff
@@ -2118,8 +2118,9 @@ public String format_csharp_contract(@GuardSatisfied ONEOFSCALAR this) {
     if (num_elts == 0) {
       return false;
     }
-
-    if (!(state instanceof INT[])) {
+    @SuppressWarnings("allcheckers:operand.instanceof.subtype") //  Warnings are suppressed because only the data type of the state needs to be checked and not the annotations
```
mernst, line 2121: You don't need to state "Warnings are suppressed because".  (That is obvious because this is a comment about the warning suppression.)

But, it would be helpful to explain *why* the qualifier is irrelevant.

java/daikon/inv/unary/OneOf.java.jpp
```diff
@@ -2118,8 +2118,9 @@ public String format_csharp_contract(@GuardSatisfied ONEOFSCALAR this) {
...
+    boolean stateTypeMatch = !(state instanceof INT[]);
```
mernst, line 2122: "stateTypeMatch" is a bad variable name because this is looking for a **mis**match.
I think "stateIsNotArray" would be a clearer variable name.

java/daikon/inv/unary/OneOf.java.jpp
```diff
@@ -2118,8 +2118,9 @@ public String format_csharp_contract(@GuardSatisfied ONEOFSCALAR this) {
     if (num_elts == 0) {
       return false;
     }
-
-    if (!(state instanceof INTARRAY)) {
+    @SuppressWarnings("allcheckers:instanceof.unsafe") //  Only the data type of the state needs to be checked and the annotations must be ignored, hence, the qualifier is irrelevant
```
mernst, line 2121: Are you sure that this warning is still issued here?  I have not been able to reproduce the warning, using your branch.  Maybe I am missing something.  Could you double-check?

aditya3434: I removed the @SuppressWarnings and the error was not raised. I guess the new commits somehow fixed the issue. So I think this PR is no longer required.

mernst: @aditya3434 If https://github.com/codespecs/daikon/pull/272 is no longer required, please close it.

PR #289, opened 2020-11-18
Remove support for the obsolete Version 1 file format

markro49:


java/daikon/chicory/DeclReader.java
```diff
@@ -296,22 +219,34 @@ public void read(File pathname) throws IOException {
...
+    String firstLine = decl_file.readLine();
     // Read the name of the program point
-    String pptname = decl_file.readLine();
-    if (pptname == null) throw new Error("file " + decl_file.getFileName() + " ends prematurely");
+    String pptname = firstLine.substring(4); // skip "ppt "
     assert pptname.contains(":::");
     DeclPpt ppt = new DeclPpt(pptname);
     ppts.put(pptname, ppt);
 
+    // Skip the ppt-type record.
```
mernst, line 237: What about `flags` and `parent` as described at http://plse.cs.washington.edu/daikon/download/doc/developer.html#Program-point-declarations ?

mernst (changes requested):

java/daikon/chicory/DeclWriter.java
```diff
@@ -714,6 +480,115 @@ private String out_name(Enum<?> e) {
     return e.name().toLowerCase();
   }
 
+  /**
+   * Output most of the decl file information for a single variable.
```
mernst, line 484: Regarding "most", what information is not output?

Also, in what the format is the information output?  Is it output to stdout, or to somewhere else?

java/daikon/chicory/DeclWriter.java
```diff
@@ -714,6 +480,115 @@ private String out_name(Enum<?> e) {
...
+   *
+   * @param parent parent of var in the variable tree
+   * @param var variable whose values are to be output
+   * @param compare_ppt ppt with compare value if comparability-file present, null otherwise
+   * @param callback object containing getComparability method
+   */
+  public void print_decl(
```
mernst, line 491: It's considered good Java style to use camelCase rather than snake_case for identifiers.  We don't have to go back and change all the old ones, but let's use camelCase for any newly-introduced identifiers.  I see you made some changes like this, such as changing print_decl_class to printDeclClass (thanks!), but you introduced other variable names in snake_case, such as creating a new print_decl method in that same class.  Could you change them?

mernst (changes requested):

mernst (changes requested):

java/daikon/chicory/DeclWriter.java
```diff
@@ -77,25 +75,17 @@ public DeclWriter(PrintStream writer) {
...
-      outFile.printf("decl-version 2.0%n%n");
-      outFile.printf("var-comparability %s%n%n", comparability);
-    } else {
-      outFile.printf("VarComparability%n%s%n%n", Runtime.comp_info != null ? "implicit" : "none");
-
-      outFile.println("ListImplementors");
-      outFile.println("java.util.List");
-      outFile.println();
-    }
+    outFile.printf("decl-version 2.0%n");
```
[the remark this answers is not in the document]
markro49: Just to make the headers on the files output by DynComp and Chicory have a similar format.

java/daikon/chicory/DeclReader.java
markro49: Many of the record types are not needed when all we care about is comparability values.  I have updated a couple of the comments to reflect this.

java/daikon/chicory/DeclWriter.java
markro49: I have updated the comment header:
    * Output most of the decl file information for a single variable.
   * This includes the variable, var-kind, enclosing-var, array, dec-type, rep-type, constant, function-args, flags, and
   * comparability records. Most notably, it does not output the parent record.  The records are output via the PrinterWriter
   * passed as an argument to the DeclWriter constructor.

java/daikon/dcomp/DCRuntime.java
```diff
@@ -98,6 +107,17 @@
...
 
+  /** Decl writer to share output code with Chicory. */
+  // Set in Premain.premain().
+  static DeclWriter decl_writer;
+
+  /** Instance of DCRuntime to use for IComparability callbacks. */
+  // Set in Premain.premain().
+  static DCRuntime runtime_object;
+
+  /** Flag that indicates when the first class has been processed. */
```
mernst, line 118: Actually this seems to indicate when the first class has *not* been processed.
Can you improve its name and documentation?

Also, please don't introduce *new* identifiers in snake_case.

mernst (changes requested):

mernst (approved):

PR #299, opened 2020-11-24
Add Javadoc

mernst:


markro49:
line 56 of DeclReader: “ture” ->”true”.

PR #313, opened 2021-02-05
Change daikon regression tests so that Chicory uses DynComp comparability file

markro49:


tests/Makefile.common
```diff
@@ -537,8 +536,9 @@ $(PROJECT_FILE).decls-DynComp $(PROJECT_FILE).txt-cset \
   $(PROJECT_FILE).decls-DynComp-nojdk $(PROJECT_FILE).txt-cset-nojdk: $(DCOMP_PREMAIN) $(DCOMP_RT) \
 				$(SCRATCH_FOR_DAIKON)/.copy-timestamp $(INST_LIST_FILE)
 	@echo
-	@echo "### Run DynComp (${PROJECT_FILE})"
+	@echo "### Compile source and run DynComp (${PROJECT_FILE})"
 	@echo
+	@echo inst_Files: '$(INST_FILES)'
```
mernst, line 541: Was this for debugging, or should it be retained?

tests/Makefile.common
```diff
@@ -855,14 +855,14 @@ ifdef ESC_ON
 ifndef NO_MERGE
 ifndef NO_CSHARP
 # ESC  MERGE  CSHARP
-txt-diff: txt-daikon-diff txt-chicory-diff txt-esc-diff inv-checker-diff \
-		  txt-jml-diff txt-merge-diff spinfo-static-diff txt-daikon-end2end-diff txt-chicory-diff \
-		  dyncomp-diff txt-simplify-diff txt-csharpcontract-diff txt-jaif-diff
+txt-diff: dyncomp-diff txt-daikon-diff txt-chicory-diff txt-esc-diff inv-checker-diff \
```
mernst, line 858: I see why dyncomp-diff comes early:  if it differs, many other things are likely to differ too and so it should be shown first.

Why is txt-chicory-diff removed?  I think it could possibly change without affecting the Daikon output, and that would be good for tests to notice.

mernst (approved):

mernst:
Thanks for this improvement to the tests.

tests/Makefile.common
markro49: These diff targets have a prerequisite chain that leads all the way back to running the requisite tool.  Putting dyncomp-diff first ensures that DynComp is run prior to Chicory.  The prerequisite 'txt-chicory-diff' was duplicated so I removed the second instance.

tests/Makefile.common
markro49: I just moved this line, along with the compilation of the test case, from the Chicory recipe to the DynComp one, but we could remove it if you wish.

markro49: i commented out the line

PR #317, opened 2021-02-23
change JDK11 dcomp-rt build to be similar to JDK 8

markro49:
(not ready for review)


java/daikon/dcomp-dummy/DCRuntime.java.dummy
```diff
@@ -0,0 +1,160 @@
+package java.lang;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+/**
+ * Runtime support for DynComp, a comparability front end for Chicory. This class is a collection of
```
mernst, line 6: I think this class needs to contain exactly the same methods and signatures as the real DCRuntime.  Could you document that correspondence both here and in the real DCRuntime?  Is there any way that a Daikon developer can avoid letting discrepancies creep in, such as extra methods here?

java/daikon/dcomp-dummy/README
```diff
@@ -0,0 +1,36 @@
+This directory contains the initial, dummy version of java.lang.DCRuntime.
+
+Post Java 8 there are increased security checks when loading JDK classes. In particular, the
```
mernst, line 3: This text appears redundantly in 3 places.  Please retain the text in just one place, and have the others point to the canonical explanation.

java/daikon/dcomp-transfer/DCRuntime.java
```diff
@@ -0,0 +1,204 @@
+package java.lang;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+/**
+ * Runtime support for DynComp, a comparability front end for Chicory. This class is a collection of
```
mernst, line 6: There is another version of this file (`Daikon.java.dummy`).  Can one of them be deleted?  If both are necessary, can one be generated from the other to avoid redundancy and the possibility of them falling out of sync?

mernst (approved):

java/daikon/dcomp-transfer/DCRuntime.java
markro49: I doubt these two derived versions of DCRuntime.java will ever need to be modified.  Hence, I don't think we need to create such a tool or script.  On the slim chance that a new method does need to be added, the README in daikon/dcomp explains how to do so.

java/daikon/dcomp-dummy/README
markro49: Consolidated the information in dcomp/README.  The two READMEs in dcomp-dummy and dcomp-transfer now just point to this master.

PR #323, opened 2021-03-25
add some baselines I missed

markro49:


markro49:
need to use a branch not master

Issue #340, opened 2021-05-19
DynComp: a fatal error has been detected by the Java Runtime Environment


boyang9602:
Hi,

I met an error when running DynComp on the Math project from defects4j. When I use the [TestRunner](https://github.com/boyang9602/APR_resources/blob/master/scripts/launcher/TestRunner.java), some of the bug ids will have the issue https://github.com/codespecs/daikon/issues/304, some of which could be solved by Mark’s solution. However, some others will have
```
.#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00007f390c7d91c2, pid=22056, tid=0x00007f3923286700
#
# JRE version: OpenJDK Runtime Environment (8.0_292-b10) (build 1.8.0_292-8u292-b10-0ubuntu1~18.04-b10)
# Java VM: OpenJDK 64-Bit Server VM (25.292-b10 mixed mode linux-amd64 )
# Problematic frame:
# j  org.apache.commons.math3.util.FastMathLiteralArrays.<clinit>()V+1377
#
# Failed to write core dump. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
# An error report file with more information is saved as:
# /home/ubuntu/APR_Projects/Math/2/Math2f/hs_err_pid22056.log
#
# If you would like to submit a bug report, please visit:
#   http://bugreport.java.com/bugreport/crash.jsp
#
```

```
ubuntu@repair:~$ java -version
openjdk version "1.8.0_292"
OpenJDK Runtime Environment (build 1.8.0_292-8u292-b10-0ubuntu1~18.04-b10)
OpenJDK 64-Bit Server VM (build 25.292-b10, mixed mode)
```

I used the Daikon https://github.com/codespecs/daikon/pull/338 cloned from the github.

The detailed reproducing steps are:
1. install defects4j, https://github.com/rjust/defects4j 
2. checkout the Math-2f using `defects4j checkout -p Math -v 2f -w Math2f`
3. go to Math2f and compile the project using `defects4j compile`
4. write the code at the end to TestIt.java and compile it
5. go to Math2f and run `java -Xmx43G -cp $DAIKONDIR/daikon.jar:target/classes/:target/test-classes/:<the helper's path> daikon.DynComp '--ppt-select-pattern=^org\.apache\.commons\.math3\.distribution\.HypergeometricDistribution' org.junit.runner.JUnitCore TestIt`
```
[...]

PR #343, opened 2021-05-27
Check for <clinit> too large to instrument

markro49:
Daikon issue https://github.com/codespecs/daikon/issues/340


mernst (approved):
Thanks!
I simplified the assignment to the new boolean; since there are no tests, could you double-check that before squash-and-merging?

markro49:
verified still fixes problem

PR #349, opened 2021-06-08
Don't use `Character.getNumericValue`

mernst:


markro49:
I'm not familiar with this code.  Is this fixing a bug?  Because the two statements are not equivalent.

mernst:
I can't figure out why this was ever using Character.getNumericValue.  That method has crazy semantics; for instance, given "Ⅼ" ('\u216C', the roman numeral fifty) it returns the value 50.  Use of `getNumericValue` seems like a bug to me.

markro49 (approved):

mernst:
I experimentally replaced the line by `throw new Error()` and all tests passed.  So I don't know whether this *can* get executed.

PR #350, opened 2021-06-08
Move instructions into a Makefile

mernst:


markro49:
I cloned and tried the make and looks good.  I have a question about:
usejdk8 && \
make
This makes 'usejdk8' look like a script or program.  Maybe you have such a thing, but I don't think others will.

mernst:
@markro I eliminated the use of my `usejdk8` shell function.  This seems to work for me now; could you double-check?  Thanks!

markro49:
I tried both java 8 and java 11 - got the correct results.

markro49 (approved):

PR #352, opened 2021-06-09
Remove `daikon.Runtime` class

mernst:
We don't know what it is for, and it does not seem to be used.


markro49:
What about daikon/java/daikon/test/split/targets/fuzzy.spinfo?  Seems to reference daikon.Runtime.ps_count

mernst:
I don't know what the fuzzy.spinfo file is.  It doesn't seem to be used. :-(

markro49 (approved):

markro49:
This pull request obsoletes 348?

mernst:
> This pull request obsoletes #348?

Yes, it does.  It was inspired by your comment on #348.

PR #358, opened 2021-07-18
New clone methods

markro49:
Changes to DCInstrument and DCRuntime to do user calls to clone in runtime and avoid introducing new branches and StackMaps in instrumented code.

Update bcel-util.jar to get improvements and corrections to discovering local variables and compiler generated temporaries.


java/daikon/dcomp/DCRuntime.java
```diff
@@ -1135,7 +1148,8 @@ public static void enter(Object[] tag_frame, @Nullable Object obj, int mi_index,
       Throwable stack = new Throwable("enter");
       StackTraceElement[] ste_arr = stack.getStackTrace();
       StackTraceElement ste = ste_arr[1];
-      if (ignore_toString && ste.getMethodName().equals("toString")) {
+      // Simplifies printouts for debugging if we ignore toString.
+      if (ste.getMethodName().equals("toString")) {
```
mernst, line 1152: Does this logic change affect behavior?  When `debug` is true and the method is `toString`, the method returns early, not executing any of the code after the `if (debug)` conditional.  That seems wrong to me; am I missing something?

The same comment applies to the hunk below.

mernst (changes requested):
Overall, his looks good.  I have some questions, especially about the use of static variables.

java/daikon/dcomp/DCRuntime.java
```diff
@@ -393,180 +392,194 @@ public static boolean dcomp_super_equals(Object o1, Object o2) {
...
-      }
-    } catch (IllegalAccessException e) {
-      throw new RuntimeException(e);
-    } catch (NoSuchMethodException e) {
-      throw new RuntimeException("error finding clone in " + target_class.getName(), e);
-    } catch (InvocationTargetException e) {
-      throw e.getCause();
+    if (clone_obj == null) {
+      throw dcomp_clone_error;
     }
```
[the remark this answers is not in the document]
markro49: The problem is we need to unwind the active_clone_calls map.  I could pass a flag to dcomp_clone_worker to indicate whether called from dcomp_clone or dcomp_super_clone.  If we did that I think we could get rid of dcomp_clone_error and have dcomp_worker_throw.  What do you think?

java/daikon/dcomp/DCRuntime.java
markro49: I ran an experiment and disabled the check.  Did not seem to make any difference in the debug output so I will remove the code.

PR #376, opened 2021-12-29
test daikon on azure with jdk17

markro49:


markro49:
Notes:
- fedora 17 all fail with  Error pulling image mdernst/daikon-fedora-jdk17
- other than that circleci passes
- on azure the typechecker run fails with
javac -cp /__w/1/checker-framework/checker/dist/checker.jar:/__w/1/s/java/lib/*:/__w/1/s/java -Xlint -J-Xmx3600m -g -XDignore.symbol.file -Xmaxerrs 100000 -Xmaxwarns 100000 -source 8 -target 8 -Werror -Xlint:-path -Xlint:-options -Xlint:-classfile -processor org.checkerframework.checker.formatter.FormatterChecker -implicit:class -Xlint:-processing -AskipDefs=^jtb\. -Xmaxerrs 10000 -Xmaxwarns 10000 -ArequirePrefixInWarningSuppressions -AwarnUnneededSuppressions -XDignore.symbol.file -Xlint:deprecation -proc:only -processorpath /__w/1/checker-framework/checker/dist/checker.jar -J--add-opens=jdk.compiler/com.sun.tools.javac.comp=ALL-UNNAMED ...


An annotation processor threw an uncaught exception.
Consult the following stack trace for details.
java.lang.IllegalAccessError: class org.checkerframework.javacutil.AbstractTypeProcessor (in unnamed module @0x56f4468b) cannot access class com.sun.tools.javac.processing.JavacProcessingEnvironment (in module jdk.compiler) because module jdk.compiler does not export com.sun.tools.javac.processing to unnamed module @0x56f4468b
	at org.checkerframework.javacutil.AbstractTypeProcessor.init(AbstractTypeProcessor.java:97)
	at org.checkerframework.framework.source.SourceChecker.init(SourceChecker.java:532)
	at jdk.compiler/com.sun.tools.javac.processing.JavacProcessingEnvironment$ProcessorState.<init>(JavacProcessingEnvironment.java:701)
	at jdk.compiler/com.sun.tools.javac.processing.JavacProcessingEnvironment$DiscoveredProcessors$ProcessorStateIterator.next(JavacProcessingEnvironment.java:828)
	at jdk.compiler/com.sun.tools.javac.processing.JavacProcessingEnvironment.discoverAndRunProcs(JavacProcessingEnvironment.java:924)
	at jdk.compiler/com.sun.tools.javac.processing.JavacProcessingEnvironment$Round.run(JavacProcessingEnvironment.java:1267)
	at jdk.compiler/com.sun.tools.javac.processing.JavacProcessingEnvironment.doProcessing(JavacProcessingEnvironment.java:1382)
	at jdk.compiler/com.sun.tools.javac.main.JavaCompiler.processAnnotations(JavaCompiler.java:1234)
	at jdk.compiler/com.sun.tools.javac.main.JavaCompiler.compile(JavaCompiler.java:916)
	at jdk.compiler/com.sun.tools.javac.main.Main.compile(Main.java:317)
	at jdk.compiler/com.sun.tools.javac.main.Main.compile(Main.java:176)
	at jdk.compiler/com.sun.tools.javac.Main.compile(Main.java:64)
	at jdk.compiler/com.sun.tools.javac.Main.main(Main.java:50)
make: *** [Makefile:941: typecheck-formatter] Error 3

java/daikon/VarInfoName.java
```diff
@@ -1394,8 +1394,8 @@ protected String java_family_name(OutputFormat format, VarInfo v) {
         String term_name = null;
         if (format == OutputFormat.JML) {
           term_name = term.jml_name(v);
-        } else if (format == OutputFormat.JAVA) {
-          term_name = term.java_name(v);
+          // } else if (format == OutputFormat.JAVA) {
```
mernst, line 1397: What is the reason for this change?

java/Makefile
```diff
@@ -86,6 +87,7 @@ else
   JAVA_RELEASE_NUMBER := $(shell echo $(JAVA_VERSION_STRING) | sed 's/\([0-9]*\)\..*/\1/')
   ifeq ($(shell expr $(JAVA_RELEASE_NUMBER) \>= 11),1)
 	DOCLET_FLAGS := --add-exports jdk.compiler/com.sun.tools.javac.code=ALL-UNNAMED
+	LINKOFFLINE := https://docs.oracle.com/en/java/javase/11/docs/api/
```
mernst, line 90: 1. I would prefer to use the latest long-term support version, which is 17.
2. Is it possible to use version 17 even when compiling under Java 8 or Java 11?  I would think so, and I think that would be nicest.

Then, this particular change (and the related changes on line 1071/1081) should be pull-requested separately, since it is a good change but has nothing to do with the main point of the pull request.

mernst (changes requested):

mernst:
1. I think I fixed the fedora-jdk17 problems.
2. That java.lang.IllegalAccessError looks to me like the problems that we added all the `-J--add-exports=` to fix.  Is it possible that those command-line argument are needed elsewhere in addition to where you have currently added them?

java/daikon/VarInfoName.java
markro49: If you look up to line 1392 you'll see that the only way to get to the code in question is if format != OutputFormat.JAVA.  This is the new issue errorprone found.

mernst: I see.  Thanks.  In that case, I would just omit it.  Leaving it commented, without an explanation, is confusing.  Thanks.

mernst: Better, the deletion could be done in a different pull request, since it isn't _really_ related to Java 17.

java/Makefile
markro49: The java17 build will fail without this change.  I seem to remember it was module related.  Java 8 doesn't have them and Java 11 does.  So I can try changing to java 17 docs, but I think it should be part of this pull request.

java/daikon/VarInfoName.java
markro49: If we move this change to a different pull request then the current pull request will fail the error-prone test.  I will delete the two lines and leave the change in for now.

markro49:
Adding the -J--add-exports fixed the checker-framework problem.  I think I will make another commit cleaning up the Makefile.

java/Makefile
markro49: changed linkoffline to jdk 17 docs

markro49:
Think all is ready for final review.

mernst (approved):
Thanks!

PR #389, opened 2022-06-27
Clean up Daikon --show-progress output.

markro49:
Reduce the default amount of output.
Give all timing output similar format.
Add new option "show-detail-progress" to get full progress output.


java/daikon/Daikon.java
```diff
@@ -287,6 +287,7 @@ private Daikon() {
 
   // When true, show how much time each program point took.
   // Has no effect unless no_text_output is true.
+  // TODO: The statement above makes sense, but is not true.  (markro 6/9/2022)
```
mernst, line 290: Can you say what *is* true?

java/daikon/LogHelper.java
```diff
@@ -22,6 +22,7 @@ private LogHelper() {
 
   // Class variables so user doesn't have to use "Level." prefix.
   public static final Level FINE = Level.FINE;
+  public static final Level FINER = Level.FINER;
```
mernst, line 25: Rather than defining all these variables, it would be possible to use `static import` to permit them to be used without a prefix in this file.  Within other files, I don't see the benefit of using `LogHelper.FINER` vs. `Level.FINER`.

java/daikon/LogHelper.java
```diff
@@ -22,6 +22,7 @@ private LogHelper() {
 
   // Class variables so user doesn't have to use "Level." prefix.
```
mernst, line 23: Who is the user?  The programmer editing this file, or a user of Daikon, or someone else?

mernst (approved):
Thanks for the cleanup.

java/daikon/LogHelper.java
markro49: done

java/daikon/LogHelper.java
markro49: code was deleted as part of previous comment resolution

mernst (approved):

PR #411, opened 2022-08-02
Add JUnit5 support to DynComp

markro49:


java/daikon/dcomp/DCInstrument.java
```diff
@@ -1811,6 +1842,51 @@ void add_exit(MethodGen mg, MethodInfo mi, int method_info_index) {
...
+   * @param targetClass JavaClass to search
+   * @param method_name target method to search for
+   * @param arg_types the target method's argument types
+   * @return name of class containing target method or null if not found
+   */
+  @ClassGetName String search_interfaces(JavaClass targetClass, String method_name, Type[] arg_types) {
+
+    // debug code
+    // System.out.println("search_interfaces of: " + targetClass.getClassName());
+    for (String interfac_temp : targetClass.getInterfaceNames()) {
```
mernst, line 1857: Try puting `@SuppressWarnings("signature:assignment")
      @ClassGetName` after the open parenthesis.

java/daikon/DynComp.java
```diff
@@ -238,11 +234,12 @@ void start_target(String premain_args, String[] target_args) {
...
     }
 
+    @SuppressWarnings("regex:assignment")
+    @Regex String file_separator = System.getProperty("file.separator");
     // If not on the classpath look in ${DAIKONDIR}/java
     String daikon_dir = System.getenv("DAIKONDIR");
     if (premain == null) {
       if (daikon_dir != null) {
-        String file_separator = System.getProperty("file.separator");
         File poss_premain = new File(daikon_dir + file_separator + "java", "dcomp_premain.jar");
```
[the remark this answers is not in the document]
markro49: I used File.separator instead.  Deleted file_separator.

java/daikon/DynComp.java
```diff
@@ -307,6 +303,20 @@ void start_target(String premain_args, String[] target_args) {
       }
     }
 
+    // For some unknown reason, groovy cannot be on the boot classpath.
+    String[] cpaths = cp.split(path_separator);
+    for (int i = 0; i < cpaths.length; i++) {
+      if (cpaths[i].contains("groovy")) {
+        String[] path_elements = cpaths[i].split(file_separator);
+        if (path_elements[path_elements.length - 1].matches("^groovy.*\\.jar$")) {
+          cpaths[i] = "";
```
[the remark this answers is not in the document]
markro49: Yes, but seemed easier than compressing the cpaths array.

java/daikon/dcomp/DCInstrument.java
```diff
@@ -426,8 +427,14 @@ public JavaClass instrument() {
...
-          if (stack_trace[i].getClassName().contains("JUnitCommandLineParseResult")
-              && stack_trace[i].getMethodName().equals("parse")) {
+          // debug code
+          // Instrument.debug_transform.log(stack_trace[i].getClassName() + " : " +
+          // stack_trace[i].getMethodName());
+          String st_classname = stack_trace[i].getClassName();
+          if ((st_classname.contains("JUnitCommandLineParseResult") // JUnit4
+                  && stack_trace[i].getMethodName().equals("parse"))
+              || (st_classname.contains("EngineDiscoveryRequestResolution") // JUnit5
+                  && stack_trace[i].getMethodName().equals("resolve"))) {
```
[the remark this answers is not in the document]
markro49: done

java/daikon/dcomp/DCInstrument.java
markro49: Could not get this to work as then conflict with @ClassGetName interfac_temp and @SignatureUnknown getInterfaceNames().

markro49:
I think I have addressed all your comments.  The one CI failure is due to the checker problem we discussed

java/daikon/dcomp/DCInstrument.java
```diff
@@ -410,7 +411,7 @@ public JavaClass instrument() {
     // We must also remember the class name so if we see a subsequent
     // call to one of its methods we do not add the dcomp argument.
 
-    debug_instrument.log("junit_state: %s%n", junit_state);
+    Instrument.debug_transform.log("junit_state: %s%n", junit_state);
```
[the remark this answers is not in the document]
markro49: done. i've assigned you https://github.com/codespecs/daikon/pull/424

java/daikon/dcomp/DCInstrument.java
```diff
@@ -427,8 +427,10 @@ public JavaClass instrument() {
         stack_trace = Thread.currentThread().getStackTrace();
         // [0] is getStackTrace
         for (int i = 1; i < stack_trace.length; i++) {
-          if (stack_trace[i].getClassName().contains("JUnitCommandLineParseResult")
-              && stack_trace[i].getMethodName().equals("parse")) {
+          // debug code
```
mernst, line 430: Minor: An alternative to the `// debug code` comments would be to add a new `static final boolean debug` variable.  Then, a programmer can turn all the debugging on or off by editing its assignment, and the `// debug code` comments would not be necessary.

java/daikon/dcomp/DCInstrument.java
```diff
@@ -722,6 +737,24 @@ public JavaClass instrument() {
...
   }
 
+  /**
+   * Returns true if the specified classname.method_name is the root of JUnit startup code.
+   *
+   * @param classname class to be checked
+   * @param method_name method to be checked
+   * @return true if the given method is a JUnit trigger
+   */
+  boolean is_junit_trigger(String classname, String method_name) {
```
mernst, line 747: Minor:  stylistic Java code uses camelCaseVariables.  Python and C use snake_case_variables.  I don't really care, but code is a touch easier to read if it all uses a similar convention.

java/daikon/dcomp/DCInstrument.java
```diff
@@ -1812,26 +1856,70 @@ void add_exit(MethodGen mg, MethodInfo mi, int method_info_index) {
     }
   }
 
+  /**
+   * Search the interfaces of a class for an implementation of the target method. For each interface
```
mernst, line 1860: Both here in the documentation, and also in the method name, please focus on *what* the method does rather than *how* it does it (by searching).  A following paragraph in the Javadoc can give the algorithm.  That can be useful information, but not as the first thing, and not as something that is visible to clients at call sites.

java/daikon/dcomp/DCInstrument.java
```diff
@@ -1812,26 +1856,70 @@ void add_exit(MethodGen mg, MethodInfo mi, int method_info_index) {
...
+   * found, test its methods to see if there is a match for the target method. If not found,
+   * recursively search the interfaces of each method.
+   *
+   * @param targetClass the JavaClass whose interfaces to search
+   * @param method_name target method to search for
+   * @param arg_types the target method's argument types
+   * @return name of class containing target method, or null if not found
```
mernst, line 1867: Above it said "search the *interfaces*", but here it says "name of *class*".  Please be consistent to avoid confusion.

java/daikon/dcomp/DCInstrument.java
mernst, line 1860: "For an implementation": might it also be a declaration, without a body?

java/daikon/dcomp/DCInstrument.java
```diff
@@ -1812,26 +1856,70 @@ void add_exit(MethodGen mg, MethodInfo mi, int method_info_index) {
...
+      if (found != null) {
+        // We have a match.
+        return found;
+      }
+    }
+    // nothing found
+    return null;
+  }
+
   /**
```
mernst, line 1904: "Discards": from where?  The stack?

java/daikon/dcomp/DCInstrument.java
```diff
@@ -1928,26 +2015,87 @@ InstructionList handle_invoke(InvokeInstruction invoke) {
...
-          if (method_name.equals("ordinal")) {
-            callee_instrumented = false;
-          }
+          // System.out.println("method: " + method_name);
+          // System.out.println("arg_types: " + Arrays.toString(arg_types));
+          // System.out.printf("invoke host: %s%n", gen.getClassName() + "." + mgen.getName());
+
+          @ClassGetName String targetClassname = classname;
+          mainloop:
+          while (true) {
```
mernst, line 2033: It would be helpful to document the purpose of this `while` loop.  Is its goal to set a particular variable, such as `callee_instrumented` and/or `classname`?

java/daikon/dcomp/DCInstrument.java
```diff
@@ -1928,26 +2015,87 @@ InstructionList handle_invoke(InvokeInstruction invoke) {
...
+            // Check that the class exists
+            JavaClass targetClass;
+            try {
+              targetClass = getJavaClass(targetClassname);
+            } catch (Throwable e) {
+              targetClass = null;
+            }
+            if (targetClass == null) {
+              // We cannot locate or read the .class file, better assume not instrumented.
+              // debug code
```
mernst, line 2043: The following line doesn't look like debug code, because it is a useful diagnostic and it is not commented.

java/daikon/dcomp/DCInstrument.java
```diff
@@ -1928,26 +2015,87 @@ InstructionList handle_invoke(InvokeInstruction invoke) {
...
+              break;
+            }
+
+            // Method not found; perhaps inherited from superclass.
+            // Cannot use "targetClass = targetClass.getSuperClass()" because the superclass might
+            // not have been loaded into BCEL yet.
+            if (targetClass.getSuperclassNameIndex() == 0) {
+              // The target class is Object; the search completed without finding a matching method.
+              // debug code
+              System.out.printf("Unable to locate method: %s%n%n", method_name);
```
mernst, line 2092: Also not debug code?

java/daikon/dcomp/DCInstrument.java
```diff
@@ -1928,26 +2015,87 @@ InstructionList handle_invoke(InvokeInstruction invoke) {
...
-          // to be finding the target of virtual call at run time.
-          // This is just a hack to get through PASCALI corpus.
-          String super_class = gen.getSuperclassName();
-          if (!super_class.equals("java.lang.Object") && BcelUtil.inJdk(super_class)) {
-            callee_instrumented = false;
+            // no methods match - does this class implement interfaces?
+            @ClassGetName String found;
+            try {
+              found =
+                  search_class_interfaces_for_target_method(targetClass, method_name, arg_types);
```
mernst, line 2070: Why does this method only search interfaces, rather than also superclasses (that is, all supertypes)?  Doing so could make the implementation cleaner here and eleminate the need for the `while (true)` loop (though the purpose of the code block should still be documented).

mernst (changes requested):
Thanks for these changes!  This is a good improvement.  I have some suggestions to make the code a bit clearer and easier to understand.

mernst:
@markro49 This is ready for your action

java/daikon/dcomp/DCInstrument.java
markro49: Because we need to search all the interfaces of the current class before we recurse to its super class.

java/daikon/dcomp/DCInstrument.java
markro49: good question.  I assume the target of an invoke instruction but be an implementation?

java/daikon/dcomp/DCInstrument.java
```diff
@@ -1837,9 +1869,68 @@ void add_exit(MethodGen mg, MethodInfo mi, int method_info_index) {
...
+        // We have a match.
+        return found;
+      }
+    }
+    // nothing found
+    return null;
+  }
+
+  /**
+   * Process an Invoke instruction. There are three cases:
```
mernst, line 1918: This comment about the logic of the method is helpful.  However, it does not match the structure of the implementation.  The implementation first computes `callee_instrumented` and only then tests for `Object.equals` and `Object.clone`.
I think it would be more efficient to make those tests first, and the method would be more like this documentation.

java/daikon/dcomp/DCInstrument.java
```diff
@@ -1858,17 +1949,19 @@ InstructionList handleInvoke(InvokeInstruction invoke) {
     if (invoke instanceof INVOKEDYNAMIC) {
```
mernst, line 1949: I suggest abstracting out the code from this line until the last assignment to `callee_instrumented`, into a method named something like `isCalleeInstrumented` or `isMethodInstrumented`.  This will make the main logic of the method much easier to follow, and will help to demarcate the different parts of this method.

java/daikon/dcomp/DCInstrument.java
```diff
@@ -1837,9 +1869,68 @@ void add_exit(MethodGen mg, MethodInfo mi, int method_info_index) {
...
+  /**
+   * Process an Invoke instruction. There are three cases:
+   *
+   * <ul>
+   *   <li>convert calls to Object.equals to calls to dcomp_equals or dcomp_super_equals
+   *   <li>convert calls to Object.clone to calls to dcomp_clone or dcomp_super_clone
+   *   <li>otherwise, determine whether the target of the invoke is instrumented or not (this is the
+   *       {@code callee_instrumented} variable)
+   *       <ul>
+   *         <li>If the target method is not instrumented, generate code to discard a primitive tag
```
mernst, line 1926: The code handles these cases in the other order:  "is instrumented", then "is not instrumented".  Change this or the code, to make them consistent with one another.

mernst, line 1926: "discard a primitive tag" sounds scary.  Is the information lost?  Why is it OK to lose that information?

java/daikon/dcomp/DCInstrument.java
```diff
@@ -1953,34 +2048,104 @@ InstructionList handleInvoke(InvokeInstruction invoke) {
...
         }
       }
     }
 
     if (invoke instanceof INVOKESPECIAL) {
       if (classname.equals(gen.getSuperclassName()) && method_name.equals("<init>")) {
-        constructor_is_initialized = true;
+        this.constructor_is_initialized = true;
       }
     }
```
mernst, line 2151: As noted elsewhere, I think the tests `if (is_object_equals(...))` and ``if (is_object_clone(...))` should occur before the call to `isCalleeInstrumented`, for efficiency and for consistency with the documentation.

mernst (changes requested):
Thanks!  I have some suggestions about the structure of the code.

java/daikon/dcomp/DCInstrument.java
```diff
@@ -1837,11 +1869,72 @@ void add_exit(MethodGen mg, MethodInfo mi, int method_info_index) {
...
+   * @param method_name target method to search for
+   * @param arg_types the target method's argument types
+   * @return name of interface class containing target method, or null if not found
+   */
+  @Nullable @ClassGetName String findTargetInterface(JavaClass startClass, String method_name, Type[] arg_types) {
+
+    if (debugFindTargetInterface) {
+      System.out.println("searching interfaces of: " + startClass.getClassName());
+    }
+    for (String interfaceNameString : startClass.getInterfaceNames()) {
```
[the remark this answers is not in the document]
markro49: 1) If we use getInterfaces we end up with two caches, on in BCEL proper and one in DCInstrument.  Probably could rewrite getJavaClass to only use the BCEL cache, but don't see any real gain for that work.
2) I don't think getAllInterfaces will work as it searches the super class before the interfaces of the current class.

java/daikon/dcomp/DCInstrument.java
mernst: I don't know the answer to this.  Sorry.

java/daikon/dcomp/DCInstrument.java
```diff
@@ -1837,43 +1866,226 @@ void add_exit(MethodGen mg, MethodInfo mi, int method_info_index) {
...
+      // because they do not have the dcomp_marker added to the argument list, but
+      // they actually contain instrumentation code.  So we do not want to discard
+      // the primitive tags prior to the call.
+      if (!junit_test_set.contains(classname)) {
+        il.append(discard_primitive_tags(arg_types));
+      }
+
+      // Add a tag for the return type if it is primitive
+      if ((ret_type instanceof BasicType) && (ret_type != Type.VOID)) {
+        // System.out.printf("push tag for return  type of %s%n",
```
mernst, line 2013: Should this be protected by an `if` test rather than being commented out?

java/daikon/dcomp/DCInstrument.java
```diff
@@ -1837,43 +1866,226 @@ void add_exit(MethodGen mg, MethodInfo mi, int method_info_index) {
...
-      // debug code
-      // System.out.printf("invokedynamic NOT the classname: %s%n", invoke.getClassName(pool));
+      if (debugHandleInvoke) {
+        System.out.printf("invokedynamic NOT the classname: %s%n", classname);
+      }
+      callee_instrumented = false;
+    } else if (is_object_method(method_name, invoke.getArgumentTypes(pool))) {
       callee_instrumented = false;
     } else {
       classname = invoke.getClassName(pool);
```
mernst, line 2071: This reassigns the formal parameter.  Is that intentional?  What is the relationship between the original value and the new value?

java/daikon/dcomp/DCInstrument.java
```diff
@@ -1837,43 +1866,226 @@ void add_exit(MethodGen mg, MethodInfo mi, int method_info_index) {
...
-      // debug code
-      // System.out.printf("callee_instrumented: %s%n", callee_instrumented);
-      // System.out.printf("invoke host: %s%n", gen.getClassName() + "." + mgen.getName());
-      // System.out.printf("invoke targ: %s%n", classname + "." + method_name);
+      if (debugHandleInvoke) {
+        System.out.printf("invoke host: %s%n", gen.getClassName() + "." + mgen.getName());
+        System.out.printf("invoke targ: %s%n", classname + "." + method_name);
+      }
 
       if (BcelUtil.javaVersion > 8) {
```
mernst, line 2079: Why only for Java versions greater than 8?

java/daikon/dcomp/DCInstrument.java
```diff
@@ -1837,43 +1866,226 @@ void add_exit(MethodGen mg, MethodInfo mi, int method_info_index) {
...
+   * @param method_name target method of the invoke
+   * @param arg_types argument types of target method
+   * @return true if target instrumented
+   */
+  boolean isTargetInstrumented(
+      InvokeInstruction invoke,
+      @ClassGetName String classname,
+      String method_name,
+      Type[] arg_types) {
+    boolean callee_instrumented;
```
mernst, line 2059: Minor: This variable's name is now inconsistent with the method's name.

java/daikon/dcomp/DCInstrument.java
```diff
@@ -1901,7 +2113,7 @@ InstructionList handleInvoke(InvokeInstruction invoke) {
       // could detect functional interfaces in a manner similar to the Java
       // compiler, but for now we will go with this simpler method.
       //
-      // Note that to simplfy our code we set the access flags for a functional
+      // Note that to simplify our code we set the access flags for a functional
       // interface to ANNOTATION in our class_access_map.
```
mernst, line 2115: I suggest renaming `class_access_map` to either `classToAccessFlags` or even just `accessFlags`, which is what it contains.

java/daikon/dcomp/DCInstrument.java
```diff
@@ -1953,132 +2167,115 @@ InstructionList handleInvoke(InvokeInstruction invoke) {
...
+            if (targetClass.getSuperclassNameIndex() == 0) {
+              // The target class is Object; the search completed without finding a matching method.
+              if (debugHandleInvoke) {
+                System.out.printf("Unable to locate method: %s%n%n", method_name);
+              }
+              callee_instrumented = false;
+              break;
+            }
+            // Recurse looking in the superclass.
+            @SuppressWarnings("signature:assignment")
```
mernst, line 2229: Here is another place that annotations in BCEL could eliminate the warning (or where the annotations in this code should be adjusted, I'm not sure which).

java/daikon/dcomp/DCInstrument.java
```diff
@@ -1953,132 +2140,154 @@ InstructionList handleInvoke(InvokeInstruction invoke) {
...
+            break;
+          }
+        }
+        class_access_map.put(classname, access);
+      } else {
+        // We cannot locate or read the .class file, better pretend it is an Annotation.
+        if (debugHandleInvoke) {
+          System.out.printf("Unable to locate class: %s%n", classname);
+        }
+        access = Integer_ACC_ANNOTATION;
```
mernst, line 2277: Why is this value not cached in `class_access_map`?

mernst (changes requested):
This is looking good!  Thanks very much for the improvements.

java/daikon/dcomp/DCInstrument.java
markro49: removed test

java/daikon/dcomp/DCInstrument.java
```diff
@@ -1837,43 +1866,226 @@ void add_exit(MethodGen mg, MethodInfo mi, int method_info_index) {
...
+   * @param arg_types the target method's argument types
+   * @return the name of the interface class containing target method, or null if not found
+   */
+  @Nullable @ClassGetName String getDefiningInterface(JavaClass startClass, String method_name, Type[] arg_types) {
+
+    if (debugGetDefiningInterface) {
+      System.out.println("searching interfaces of: " + startClass.getClassName());
+    }
+    for (String interfaceNameString : startClass.getInterfaceNames()) {
+      @SuppressWarnings("signature:assignment")
```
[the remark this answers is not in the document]
markro49: I will make this a separate pull request.

java/daikon/dcomp/DCInstrument.java
markro49: I will make this a separate pull request.

mernst (approved):
This looks great.  Thanks!

mernst (approved):

PR #416, opened 2022-08-14
Remove suppression obviated by Resource Leak Checker bug fix

msridhar:
https://github.com/typetools/checker-framework/pull/5263 fixes the bug in the Resource Leak Checker that necessitated this suppression


mernst:
This is a good change, but the branch needs to be named `resource-leak-identical-returns` rather than `patch-1`.  Thanks!

PR #417, opened 2022-08-16
Remove unneeded suppressions

msridhar:
No longer needed due to https://github.com/typetools/checker-framework/pull/5263


msridhar:
Now that https://github.com/typetools/checker-framework/pull/5263 has landed, this will need to land to make the Checker Framework CI green again

PR #421, opened 2022-08-24
Cache the result of `findJavaClass()`

mernst:


markro49:
delete the line:
javaClasses.put(classname, null);
from the routine

markro49 (approved):

PR #433, opened 2022-09-17
Type-check for signedness

markro49:


mernst (changes requested):
For each newly-added `@SuppressWarnings`, could you write a comment justifying it?  I did a couple of them.

Thanks!

PR #496, opened 2023-06-29
Postpend, don't prepend, JAVA_HOME to PATH

### LINKED ISSUE #493: `daikon.bashrc` Should Not Add JAVA_HOME to $PATH
On this line of the `daikon.bashrc` script, the JAVA_HOME variable is added to $PATH.

https://github.com/codespecs/daikon/blob/169f2dab1d5d214591e74785b403517b01c25846/scripts/daikon.bashrc#L61

This causes MacOS users to be unable to switch Java versions. On MacOS, the `java` executable is "special", because it automatically detects the correct executable to use based on JAVA_HOME.

Because `daikon.bashrc` overwrites $PATH to put its inferred JAVA_HOME first, the `java` bin of the inferred version will always be used, and the "special" MacOS executable will not be run, thus preventing switching Java versions.

The best solution to this is likely to only add JAVA_HOME to $PATH if not on Darwin. I'm willing to submit a quick PR for this if that solution sounds good.

NateLevin1:
Fixes #493. Without this change, MacOS users are unable to switch Java versions.

This change is more general than my proposed solution in that issue. We simply check if `$JAVA_HOME` has already been exported, and only export it if it hasn't. This fixes the issue on MacOS and will also work for users on other platforms that use a Java version manager such as [jEnv](https://www.jenv.be).

Thanks!


scripts/daikon.bashrc
```diff
@@ -58,7 +58,12 @@ if [ -d ${DAIKONDIR}/java ]; then
 fi
 export DAIKON_CLASSPATH
 
-export PATH=${DAIKONSCRIPTS}:${PLUMESCRIPTS}:${JAVA_HOME}/bin:$PATH
+# Only add JAVA_HOME to PATH if JAVA_HOME is not set; see GitHub issue #493
+if [ -z "$JAVA_HOME" ]; then
```
mernst, line 62: This looks backwards to me.  If JAVA_HOME is *not* set, then this adds `/bin` to the path.
I think that `${JAVA_HOME}/bin` should be added to the path only if JAVA_HOME *is* set.

mernst (changes requested):

scripts/daikon.bashrc
mernst: Or maybe the change should be Darwin-specific?  (I'm not sure.)

NateLevin1: You are right — this is not the best solution. There's a few options:
1. Make the change Darwin-specific
2. Instead of prepending `$JAVA_HOME/bin` to the path, we could append it.
3. Check if a `java` executable is already in the user's `$PATH`, only add `$JAVA_HOME/bin` if one isn't
Any thoughts on which one we should choose? I am leaning towards 2.

mernst: I'm open to any of these, so your proposal of <span>#</span>2 is fine with me.

NateLevin1: I've updated this PR with the second fix — please let me know if anything else is needed. Thanks!

mernst (approved):
Thanks!

PR #497, opened 2023-07-04
Add AGORA invariants

juaaloval:
Add 22 new invariants designed for the context of REST API testing. These invariants are based on an analysis of a systematically collected dataset of 40 realistic APIs from the RapidAPI platform and were used in the paper entitled “AGORA: Automated Generation of test oracles for REST APIs”, accepted for publication at ISSTA 2023. All the new invariants are disabled by default. A complete description of these invariants can be found in the header of each invariant class and here: https://github.com/isa-group/Beet/wiki/Modified-version-of-Daikon#added-invariants 

Modify computeConfidence function of certain invariants that always returned CONFIDENCE_JUSTIFIED (Range.java.jpp, Member.java.jpp, Reverse.java.jpp, SubSequence.java.jpp, SubSet.java.jpp), now these functions return CONFIDENCE_JUSTIFIED if there is at least one occurrence of the invariant during the program execution.


java/daikon/Daikon.java
```diff
@@ -149,12 +149,10 @@
 import daikon.inv.unary.sequence.SeqIndexIntLessEqual;
 import daikon.inv.unary.sequence.SeqIndexIntLessThan;
 import daikon.inv.unary.sequence.SeqIndexIntNonEqual;
-import daikon.inv.unary.string.CompleteOneOfString;
-import daikon.inv.unary.string.OneOfString;
-import daikon.inv.unary.string.PrintableString;
-import daikon.inv.unary.stringsequence.CommonStringSequence;
-import daikon.inv.unary.stringsequence.EltOneOfString;
-import daikon.inv.unary.stringsequence.OneOfStringSequence;
+import daikon.inv.unary.string.*;
```
mernst, line 152: Please don't introduce wildcard imports.  The Daikon style is to not use them.

java/typequals/vindex/VIndexAnnotatedTypeFactory.java
```diff
@@ -11,6 +11,7 @@
 import org.checkerframework.javacutil.AnnotationBuilder;
 import org.checkerframework.javacutil.TreeUtils;
 import org.checkerframework.javacutil.TypesUtils;
+import typequals.vindex.qual.VIndexTop;
```
mernst, line 14: Why is this import statement needed?

java/daikon/inv/unary/string/dates/IsTimestampYYYYMMHHThhmmssmm.java
```diff
@@ -0,0 +1,99 @@
...
+  @Override
+  public String format_using(
+      @GuardSatisfied IsTimestampYYYYMMHHThhmmssmm this, OutputFormat format) {
+    return var().name()
+        + " is Timestamp. Format: YYYY-MM-DDTHH:MM:SS.mmZ (Miliseconds are optional)";
+  }
+
+  @Override
+  public InvariantStatus check_modified(String v, int count) {
+    Pattern pattern =
```
mernst, line 71: It is inefficient to re-compile this pattern every time that `check_modified()` is called.  Please store it in a static private variable instead.
The same comment apples to other regular expressions in this pull reqest.

mernst (changes requested):

java/daikon/Daikon.java
JuanCarlosAlonsoValenzuela: Fixed in Commit https://github.com/codespecs/daikon/pull/497/commits/a1663546c73ae48608658b86a9dfbd916736698e

java/typequals/vindex/VIndexAnnotatedTypeFactory.java
JuanCarlosAlonsoValenzuela: I get an error in line 26 of that class if I remove the import.

java/daikon/inv/unary/string/dates/IsTimestampYYYYMMHHThhmmssmm.java
JuanCarlosAlonsoValenzuela: Fixed in https://github.com/codespecs/daikon/pull/497/commits/94c84606e997d498c302d61b21bceb48a7015c61

java/daikon/Daikon.java
mernst: Thanks.  If you address a comment, you can just click the "Resolve conversation" button.  Let's reserve replies for conveying more information.

mernst:
@JuanCarlosAlonsoValenzuela I apologize for having broken the Daikon build.  Everything is fixed now, as far as I can tell.  The pull request is failing because of missing documentation.

mernst (changes requested):

JuanCarlosAlonsoValenzuela:
@mernst I have addressed all the requested changes. However, the build keeps failing due to a lack of Javadoc comments in classes in which I have no performed any modifications whatsoever.

error_screenshot

mernst:
You can use a link like https://dev.azure.com/mernstdaikon/daikon/_build/results?buildId=1453&view=logs&j=c5caa13a-7690-5f09-4f3d-82e48294691c&t=861794d2-f2dc-5d44-80b1-63fb79ef3bea&l=27342 instead of an image, if you like.

The problem was the Javadoc error in `IsEmail.java`.  (There are warnings too, but they were not stopping the build; the error was the problem.)

JuanCarlosAlonsoValenzuela:
@mernst 
I have added all the missing Javadoc documentation. The build keeps failing, but I don't know how to locate the error in the trace (https://dev.azure.com/mernstdaikon/daikon/_build/results?buildId=1455&view=logs&j=c5caa13a-7690-5f09-4f3d-82e48294691c&t=861794d2-f2dc-5d44-80b1-63fb79ef3bea). 

I have not seen any line explicitly indicating the cause of the error. I only know that it must be something related to Javadoc.

java/daikon/inv/unary/stringsequence/SequenceFixedLengthString.java
```diff
@@ -0,0 +1,144 @@
...
+import org.checkerframework.checker.interning.qual.Interned;
+import org.checkerframework.checker.lock.qual.GuardSatisfied;
+import org.checkerframework.checker.nullness.qual.Nullable;
+import org.checkerframework.dataflow.qual.Pure;
+import org.checkerframework.dataflow.qual.SideEffectFree;
+import org.checkerframework.framework.qual.Unused;
+import typequals.prototype.qual.Prototype;
+
+/**
+ * Indicates that all the elements of an array of strings have a fixed length n. Prints as {@code
```
mernst, line 17: What is the semantics of SequenceFixedLengthString (and other invariants) with respect to `null` sequence elements?  Should SequenceFixedLengthString be true of ["aaa", "bbb", null, "ccc"]?
Please document the answer and test it.
Currently, the SequenceFixedLengthString code crashes when it encounters this, and other invariants may have the same problem.

mernst (changes requested):

JuanCarlosAlonsoValenzuela:
I fixed the crashes caused by arrays of strings containing null values. The build currently fails because the AGORA invariants (now enabled by default) are being reported in the test suites. I do not recall enabling these invariants.

JuanCarlosAlonsoValenzuela:
I would also like to comment that yesterday I ran into some issues while trying to configure Daikon (maybe they are my fault, but I think that I should report them just in case).

I had to clone the repository again, so I decided to use JDK 17 while following the steps in the [Daikon installation guide](https://plse.cs.washington.edu/daikon/download/doc/daikon.html#Installing-Daikon).

After compiling everything (I use WSL) and adding all the [external jars]( https://plse.cs.washington.edu/daikon/download/doc/developer.html#Using-Eclipse) (including tools.jar), I got the following errors in the IntelliJ project:
-	`jdk.javadoc.doclet does not exist` (among other Javadoc related classes).
-	In the `daikon/java/typequals/vindex/VIndexAnnotatedTypeFactory`, IntellIj reported an error in line 25, requesting me to import `VIndexTop`.
-	In line 52 of this same class, IntellIj reported that the method `leastUpperBounds` could not be found.

After spending some time trying to fix this, I decided to downgrade to JDK 8. After doing this, the first error (Javadoc) disappeared, but the other 2 were still present. 

I realized that the content of the `typequals` directory provided in the [Daikon releases]( https://plse.cs.washington.edu/daikon/download/) (the ones I used for the AGORA paper) is different to the content in the GitHub repository, so I replaced the content of the `typequals` directory with the one provided in the releases and the errors disappeared (of course, none of these changes are part of my commits). 

This was the only way I managed to successfully execute Daikon, I don’t know if someone else using JDK 17 has run into these issues or if I made some mistake in the installation process.

mernst:
@JuanCarlosAlonsoValenzuela 
In several Agora invariants, the `merge(List, PptTopLevel)` method is not defined.  This is leading to incorrect results, because they inherit an implementation that does not work correctly.  This was in turn leading to test failures.

I have enhanced Daikon to warn about missing overrides of that method.  You can see errors like this in the `misc` job:

```Exception in thread "main" java.lang.Error: daikon.inv.unary.string.FixedLengthString defines no merge method, but these fields might store state: length```

Could you please correct the invariants so that their `merge(List, PptTopLevel)` methods behave correctly?

JuanCarlosAlonsoValenzuela:
@mernst 

I have added all the merge methods for the AGORA invariants causing the new exception that you mentioned in your previous comment (missing merge method).
I also implemented the `clone()` method for all these invariants.
I still have one doubt, in your comment and in the exception message, the method is called `merge(List, PptTopLevel)`, but I think that it should be `merge(List, PptSlice)`.
On a side note, I am still having issues to install Daikon using JDK 17 and I had to use Java 8 to test the implementation of these changes (see comment below).

JuanCarlosAlonsoValenzuela:
@mernst
As I mentioned in my previous comment, I am still facing some issues to install Daikon using Java 17. The most likely scenario is that I made some mistake when following the steps, but just in case I will share the steps that I followed in case you find that I did something wrong (I have also created a VM with the results of these steps to ease replication). In the worst case, maybe this could help to improve the documentation.

It is worth mentioning that I did not run into these issues when installing a Daikon release (available at: https://plse.cs.washington.edu/daikon/download/).

The steps that I followed in a Ubuntu VM with Java 17are the following:

1.	Clone Daikon from the branch containing this PR.
2.	Create $DAIKONDIR variable in `/home/jcav/Documents/GitHub/daikon` by adding the following lines to the `~/.bashrc` file:
`export DAIKONDIR=/home/jcav/Documents/GitHub/daikon`
`source $DAIKONDIR/scripts/daikon.bashrc`
3.	Install all the requirements for compiling Daikon (specified here: https://plse.cs.washington.edu/daikon/download/doc/developer.html#Requirements-for-compiling-Daikon), more specifically, the ones in this Dockerfile: https://github.com/codespecs/daikon/blob/master/scripts/Dockerfile-ubuntu-jdk17-plus
4.	Compile Daikon and build other tools by running: `make -C $DAIKONDIR rebuild-everything`
5.	Open the project using IntelliJ idea, set directory `/daikon/java` as Sources Root and add all jars inside the `lib` directory. I did not add the `tools.jar` (as specified here https://plse.cs.washington.edu/daikon/download/doc/developer.html#Using-Eclipse) jar because it is not present since JDK 9 (https://stackoverflow.com/questions/69353634/tools-jar-does-not-come-with-jdk-installation).
6.	After these steps (and introducing the changes of my latest commits), I ran Daikon.java and found the following exceptions:
- File: `java/daikon/dcomp-transfer/DCRuntime.java`. 
    - Error: `java: package exists in another module: java.base`.
- File: `java/typequals/vindex/VIndexAnnotatedTypeFactory.java`
    - Error 1: `Line 25, Cannot find symbol class VIndexTop`. This can be solved simply by importing it: `import typequals.vindex.qual.VIndexTop`, but this import is not added by default.
    - Error 2:  `Line 52, cannot find method leastUpperBound`.
    - It is worth mentioning that, for some reason, this file is not part of the Daikon releases that can be downloaded here: https://plse.cs.washington.edu/daikon/download/ When implementing AGORA, I used the latest release, hence why I did not find this issue until now.

Of course, I did not commit these changes without testing them first. I tested them with my computer using Java 8 and the declaration and data traces generated by Beet, but I still had to remove the `/java/typequals/vindex/VIndexAnnotatedTypeFactory.java` file.

I have sent you an email containing an Ubuntu VM with the results of the aforementioned steps, the subject of the email is “Daikon VM”. Let me know if you need more details about this issue.

PR #500, opened 2023-07-13
test confidence changes

markro49:


mernst:
@juancarlosalonsovalenzuela 
1. The output contains many "this != null" invariants that should not be printed.  Can you figure out which of the 5 changes is causing that?  Then we know where to look when investigating the cause.
2. Daikon is supposed to not print any invariants for which there are no samples.  Can you provide a reproduction of Daikon producing them?

JuanCarlosAlonsoValenzuela:
> @JuanCarlosAlonsoValenzuela
> 2. Daikon is supposed to not print any invariants for which there are no samples.  Can you provide a reproduction of Daikon producing them?
@mernst I send you an email on 14th July containing some examples of this behavior, along with explanations. Did you received it correctly?

I have also tried to add a commit to this PR (merging master into this branch), but apparently I have no permissions to do it.

JuanCarlosAlonsoValenzuela:
@mernst 

According to the diffs in the test cases (e.g., lines 5942-5949 in https://dev.azure.com/mernstdaikon/daikon/_build/results?buildId=1462&view=logs&j=dd28197a-8f18-5a93-858c-425f97f453bb&t=a772a73c-99a4-5758-5ab0-837be5038148&l=5942), the invariants that disappear from the report, causing the build failure, belong to a `Member` class (e.g., `orig(size(a[])) in u[]`), the same type as the undesired invariants in the email with examples that I provided you. 

So maybe these invariants should not have been reported in the first place.

mernst (approved):
Thanks!

PR #524, opened 2023-10-17
Fix a possible resource leak in the `Session` constructor

Calvin-L:
(Discovered by https://github.com/typetools/checker-framework/pull/6241)

After opening the output stream `trace_file`, the constructor goes on to do many initialization tasks that might throw exceptions.  If that happens, the constructor exits with a `SimplifyError` and does not close `trace_file`, leaking the open output stream.  There is no way for a caller to clean it up because the partially-constructed `Session` is not accessible after the constructor throws.

This commit fixes the problem using a temporary variable with the same name.  On exception the temporary variable will be closed before the constructor exits.

This commit also makes an effort to preserve the entire exception chain on failure, which can be helpful for debugging because it preserves the entire stack trace.

### DIFF
--- a/java/daikon/simplify/Session.java
+++ b/java/daikon/simplify/Session.java
@@ -87,6 +87,7 @@ import org.checkerframework.checker.nullness.qual.Nullable;
    * this Session.
    */
   public Session() {
+    PrintStream trace_file = null;
     try {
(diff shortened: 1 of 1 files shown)


mernst:
Thanks for this fix!

PR #528, opened 2023-12-14
Add compare decls to DtraceDiff

markro49:
add compare decls to DtraceDiff and corresponding goal file changes


markro49:
I spent a lot of time investigating the nondeterminism problem with daikon-tests/Graph. The symptom is that two classes are loaded in a different order.  You can run the test case multiple times in a row and the results will vary back and forth.  I tried numerous Java loader options, adding lots of debugging code to Instrument.java and using additional features of the Java Instrumentation interface. All to no avail.  I have spent too much time on this issue so I will just leave in the testing Makefile changes that skip the verification of the Graph.txt-chicory.dtrace file

mernst:
@markro49 Are the contents of Graph.txt-chicory.dtrace the same, just in a different order?  Or does the difference in loading order cause behavioral differences?  Also, what are the classes that are loaded in a nondeterministic order?

markro49:
The results are the same, just in a different order.  The two classes that get loaded in different order are the public class Exceptions ones at the end of the file.  I think it was NoNodeException and DuplicateNodeException; but I don't remember for sure.  It was always the same two whose order differed.

mernst:
@markro49 There is some missing documentation:

```
/__w/1/s/java/daikon/tools/DtraceDiff.java:63: warning: no comment
/__w/1/s/java/daikon/tools/DtraceDiff.java:65: warning: no comment
/__w/1/s/java/daikon/tools/DtraceDiff.java:466: warning: no comment
```

tests/daikon-tests/Graph/Makefile
```diff
@@ -6,6 +6,8 @@ MAIN_CLASS:=MapQuick1.PS4TestDriver
 
 ESC_ON:=1
 
+SKIP_DTRACE_DIFF := 1
```
mernst, line 9: Please document the reason that DtraceDiff is skipped.

mernst (approved):
Thanks!

PR #534, opened 2024-02-19
Issue 532

markro49:
We should not add the entire classpath to the boot classpath.  Just the location of daikon; either daikon.jar or daikon/java:daikon/java/lib/*.


mernst (changes requested):
Thanks for the fix.

java/daikon/DynComp.java
```diff
@@ -122,6 +142,38 @@ public class DynComp {
   /** Synopsis for the DynComp command line. */
   public static final String synopsis = "daikon.DynComp [options] target [target-args]";
 
+  /**
+   * Locate the classfile file that a class was loaded from. The result is a URL for the file
+   * (starting with "file:" or "jar:file:").
+   *
+   * @param c class to look up
+   * @return URL to the classfile, or null if not found
```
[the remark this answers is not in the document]
markro49: If whereLoadedFrom is seen as a general utility method, then it could return null if the class could not be located do to the use of a custom classloader.  It cannot happen in our case.  Do we want to keep general code or custom tailor to our use?

java/daikon/DynComp.java
```diff
@@ -122,6 +142,38 @@ public class DynComp {
...
+   */
+  public static @Nullable String whereLoadedFrom(Class<?> c) {
+    if (c == null) {
```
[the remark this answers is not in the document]
markro49: Similar to above: If this is a general utility method we need to check c before calling c.getClassLoader().  Again, this cannot happen in our case.  Do we want to keep general code or custom tailor to our use?

java/daikon/DynComp.java
```diff
@@ -454,4 +484,73 @@ public String args_to_string(List<String> args) {
     }
     return str.trim();
   }
+
+  /**
+   * Search for a file on the current classpath, then in ${DAIKONDIR}/java, then in ${DAIKONDIR}.
+   * Returns an empty file name to mean use the expansion of {@code java/lib/*}. Returns null if not
```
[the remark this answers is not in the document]
markro49: I have added several comments to clarify what is going on.  However, I just realized there is a problem.  The code assumes that if we find 'java/lib/*' on the classpath we will also see 'java' on the classpath as this is the normal debugging scenario.  But if the classpath contains 'java/lib/*' followed by daikon.jar, we will not add daikon.jar to the boot classpath which is incorrect.  I will look at this some more.

java/daikon/DynComp.java
```diff
@@ -454,4 +484,73 @@ public String args_to_string(List<String> args) {
...
+          // Return empty file name as flag to use java_lib_classpath instead.
+          return new File("");
+        }
+        return poss_file;
+      }
+    }
+
+    // Not on the class path - do special case:
+    // If looking for daikon.jar and java/lib/* is on
+    // classpath, then use that. 10 is arbitary value,
```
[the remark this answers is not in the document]
markro49: As part of solution to previous comment we may need to have locateFile return an object that contains both a string and a flag.

mernst:
@markro49 Pull master into this branch when you get a chance.

java/daikon/DynComp.java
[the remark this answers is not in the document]
mernst: Given a custom classloader, wouldn't `getResource()` still work?

java/daikon/DynComp.java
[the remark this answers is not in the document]
mernst: If this is a general utility method, its specification could say that the argument cannot be null.  Then the check wouldn't be needed.
Currently, the specification doesn't say what happens for null.  Is null being the argument is the only situation where null is the result?  If so, it would simplify the specification to forbid null as an argument.
More generally, I don't see how permitting null as an argument helps clients.

markro49:
This is not the final version.  I just wanted to checkpoint the changes I made in response to Michael's review.

markro49:
I have redone the method for computing the boot classpath.  The one failure is the change in the lastest check we have already discussed.  Please take a look at the changes.

java/daikon/DynComp.java
```diff
@@ -59,7 +67,10 @@ public class DynComp {
   // Null if shouldn't do output
   public static @MonotonicNonNull File trace_file = null;
 
-  /** Controls size of the stack displayed in tracing the interactions that occurred. */
+  /**
+   * Controls size of the stack displayed in tracing the interactions that occurred. Used in {@link
+   * TagEntry}.
+   */
   @Option("Depth of call hierarchy for line tracing")
   public static int trace_line_depth = 1;
```
mernst, line 74: Just below, does setting `abridged_vars` to true admit the possibility of a name clash and a crash?

java/daikon/DynComp.java
```diff
@@ -181,13 +205,14 @@ public static void check_args(Options options, String[] target_args) {
   }
 
   /**
-   * Starts the target program with the Java agent setup to do the transforms. All Java agent
+   * Starts the target program with the Java agent set up to do the transforms. All Java agent
    * arguments are passed to it. Our classpath is passed to the new JVM.
```
mernst, line 207: Who is "Our"?  (It could by DynComp, or the program being instrumented.)

java/daikon/DynComp.java
```diff
@@ -300,21 +281,23 @@ void start_target(String premain_args, String[] target_args) {
...
+      // If any of thses cases are true, we append the path to the boot
+      // classpath. We then repeat the process with the next element of
+      // the classpath.
+      for (String path : cp.split(File.pathSeparator)) {
+        if (isDaikonOnPath(path)) {
+          daikonPath = daikonPath + File.pathSeparator + path;
         }
       }
+      // debuging
+      System.out.println("daikonPath: " + daikonPath);
```
mernst, line 282: This should be removed or protected by a flag.

java/daikon/DynComp.java
```diff
@@ -453,4 +421,99 @@ public String args_to_string(List<String> args) {
...
+        } else {
+          poss_file = new File(path, fileName);
+        }
+      } else {
+        poss_file = new File(path, fileName);
+      }
+      if (poss_file.canRead()) {
+        return poss_file;
+      }
+    }
```
mernst, line 506: Could you abstract this block of code into a `findOnClasspath` routine?

mernst (changes requested):
Thanks for these improvements!  The code looks correct to me now.  I have a few comments.

java/daikon/DynComp.java
markro49: I don't think so.  abridged_vars is only used to print simplfied output; internal variable is unchanged.

java/daikon/DynComp.java
markro49: Changed to "The current classpath is passed ..."

java/daikon/DynComp.java
markro49: cahnged to basic.log ...

PR #539, opened 2024-04-24
Prepend `${HOME}` to Daikon installation location in docs

jyoo980:
@mernst I needed to make this change to be able to run Daikon (play-testing for CSE P 504).

### DIFF
--- a/doc/daikon.texinfo
+++ b/doc/daikon.texinfo
@@ -264,7 +264,7 @@ We will assume that you are using the bash shell or one of its variants.
 Add commands like these to your @file{~/.bashrc} or @file{~/.bash_profile} file:
 @example
 # The absolute pathname of the directory that contains Daikon
-export DAIKONDIR=@var{daikonparent}/daikon-5.8.19
+export DAIKONDIR=$@{HOME@}/@var{daikonparent}/daikon-5.8.19
 source $DAIKONDIR/scripts/daikon.bashrc
 @end example
 


doc/daikon.texinfo
```diff
@@ -264,7 +264,7 @@ We will assume that you are using the bash shell or one of its variants.
 Add commands like these to your @file{~/.bashrc} or @file{~/.bash_profile} file:
 @example
 # The absolute pathname of the directory that contains Daikon
-export DAIKONDIR=@var{daikonparent}/daikon-5.8.19
+export DAIKONDIR=$@{HOME@}/@var{daikonparent}/daikon-5.8.19
```
mernst, line 267: We don't know that Daikon is installed under home.  Does 4bfe8ef9f0b5c11eee029b6a4f327a24bc1def38 help?  Maybe it's still unclear whether directory `daikon-5.8.19` contains Daikon or is Daikon.  (I'm thinking of it as the latter.)  Maybe we should just say it's an absolute path and leave it at that?

jyoo980: OK, saying it's an absolute path should disambiguate issues that people might have while installing Daikon.

I was trying to bring the Daikon docs in-line with the docs for the Checker Framework, which appear to assume that the user has a local installation of the CF under `$HOME` [here](https://checkerframework.org/manual/#javac-wrapper)

mernst: Ah, but those CF instructions are prefixed by "For example".  :smiley:  I agree it's all just a bit imprecise, but I also hope that this won't be what trips up users.

jyoo980: I'm sure users will be able to figure it out. I'm good to merge this, if you're ready!

jyoo980:
No need to merge; the change is already on master.

PR #540, opened 2024-04-27
Simplify calls to `List.toArray()`

mernst:


DvonHolten:
just had a quick look at the changes. did that to my codebase long time ago. however, small improvement is possible: instead of creating a `new xxx[ 0 ]`  all the time, its sufficient to re-use a `final static xxx[ 0 ]` laying around somewhere

mernst:
@DvonHolten Thanks for pointing that out.  I will make that improvement.

PR #552, opened 2024-06-29
changes to run daikon on jdk21

markro49:
I think I may have messed up mernst/jdk21 so this is a copy for testing


markro49:
This should be identical to the jdk21 branch of mernst/daikon but i may have missed something.

mernst:
> This should be identical to the jdk21 branch of mernst/daikon but i may have missed something.

Yes, they are identical.

java/Makefile
```diff
@@ -30,8 +31,11 @@ else
   JAVA_RELEASE_NUMBER := $(shell echo $(JAVA_VERSION_STRING) | sed 's/\([0-9]*\)\..*/\1/')
 endif
 
+# Problem: Release 21-ea does not contain the "this-escape" lint flag.
```
mernst, line 34: I would expect that we never use 21-ea, given that JDK 21 has been released.  Do you know where we use it?

tests/kvasir-tests/Makefile.common
```diff
@@ -121,6 +121,9 @@ endif
 endif
 
 ifeq ($(OS_ID),Ubuntu)
+ifeq ($(OS_REL),23.10)
+	GOAL := goal-Ubuntu2310
```
mernst, line 125: Since Ubuntu 24.04 has been released (and it's LTS), should we just ensure that we never run tests under anything between 22.04 (the last LTS Ubuntu release) and 24.04?  That would simplify the logic here.  I would need to make new Docker images, but that is not a problem.

mernst (approved):
This looks great.  Thank you very much.
I have just a couple of comments -- which we could address after merging this if you prefer.

PR #553, opened 2024-07-05
new goal files for Ubuntu 24.04

markro49:


mernst:
I believe I need to make new Docker images to make these tests pass.  Is that right?

markro49:
They ci checks should pass as the new goal files are not being used as yet. I ran the tests on a 24.04 wsl box.

tests/kvasir-tests/Makefile.common
```diff
@@ -121,6 +121,9 @@ endif
 endif
 
 ifeq ($(OS_ID),Ubuntu)
+ifeq ($(OS_REL),24.04)
+	GOAL := goal-Ubuntu2404
+endif
 ifeq ($(OS_REL),23.10)
```
mernst, line 127: Should this and some other `ifeq` statements be removed, since the test for "24.04" now exists?

mernst (approved):
Thanks!

tests/kvasir-tests/Makefile.common
markro49: A few goal files differ between Ubuntu 22.04 and 24.04; so I think we need to keep this change for now.  We can review again when we decide what to do about the other, stale goal files.

mernst (approved):
Thanks!

PR #559, opened 2024-07-19
Permit array values in SampleTester

mernst:


markro49:
Looks ok to me.  Minor note: I tried 'make junit' and it fails as it depends on 'daikon' . Probably not worth worrying about.

PR #570, opened 2024-11-16
update goal files for Ubuntu-24.10

markro49:


mernst:
Should these changes be integrated into #569 rather than being a separate pull request?

markro49:
not a bad idea

mernst (approved):

PR #572, opened 2024-11-24
Simplify goal_name computation

mernst:


markro49 (approved):
I'm curious about changes from $() to ${}.  According to GNU make documentation there is no difference.  Is this a style choice or perhaps something to do with escaping in bash commands or ?

mernst:
I used `${...}` because it can be pasted into a bash shell whereas `$(...)` cannot.  I should have made those changes a separate pull request.

PR #574, opened 2025-01-17
Use Error Prone 2.31

markro49:
replacing error-prone version 2.15.0 with version 2.31.0 which is the last version of error-prone that will run on jdk 11.


markro49:
I got frustrated with adding all the javadoc comments and took some short cuts.  There are several places with comments of '???' where I didn't take the time to figure out a meaningful comment.

mernst:
> I got frustrated with adding all the javadoc comments and took some short cuts. There are several places with comments of '???' where I didn't take the time to figure out a meaningful comment.

When doing a widescale refactoring that would require massive documentation, you can create a file named `SKIP-REQUIRE-JAVADOC` in the top level to skip the require-javadoc test.

mernst:
Furthermore, if a warning should be suppressed throughout the codebase, you can add `-Xep:PatternMatchingInstanceof:OFF` in the Makefile and avoid all the require-javadoc warnings.

mernst (approved):
Thanks!

PR #582, opened 2025-01-25
Use latest Error Prone if java version is 17 or greater.

markro49:
Source corrections for a new warning: unneeded 'continue
and a new error:  if branches contain identical code.

I commented out the continues rather than deleting them; feel free to delete if you wish.

The identical code branches were in a .jpp file where there was a '#if' within an if statement.   Also tried to just comment out, but wasn't sure if:
return "%var1% == %var2%*%var2%";
was the same as:
return "%var1% == %var2%**2";


mernst (approved):
Thanks!

PR #583, opened 2025-01-27
remove unused statement that causes warning with newer python3

markro49:
LWArrayRExp doesn't appear to be used  and the re.compile expression gets a warning with newer versions of Python3.  Might be the case that the comment just prior to the deleted statement is no longer needed.


mernst (approved):
Thanks!

PR #585, opened 2025-02-05
Add additional version of Chicory for Java 24

markro49:
Uses new java.lang.classfile api


java/Makefile
```diff
@@ -745,17 +802,22 @@ else
...
+	  ${extra-error-prone-arg-1} \
 	  ${JAVA_FILES_FOR_STYLE}
 endif
 
+# THIS IS BROKEN - IS IT EVER USED?
 # Create an `error-prone.patch` file.
 errorprone-replacements:
 	javac -cp ${DAIKON_CLASSPATH} \
+ifeq ($(shell test ${JAVA_RELEASE_NUMBER} -gt 23; echo $$?),0)
+	  -source 24 -target 24 --enable-preview \
```
mernst, line 819: Is this the same as `${extra-error-prone-args-2}`?  That is, could you use that variable?

java/Makefile
```diff
@@ -1168,13 +1279,20 @@ DAIKON_PACKAGES := $(shell ls -F daikon | grep / | egrep -v 'util' | egrep -v 'd
...
-	find * \( -name 'PrototypeChecker.java' -o -name 'VIndexChecker.java' -o -name 'VIndexAnnotatedTypeFactory.java' -o -path 'daikon/dcomp-transfer' \) -prune -o -name "*.java" -print | ${SORT_DIRECTORY_ORDER} > javadoc-files.txt
+	find * \( -name 'PrototypeChecker.java' -o -name 'VIndexChecker.java' -o -name 'VIndexAnnotatedTypeFactory.java' -o -path 'daikon/dcomp-transfer' \) -prune -o -name "*24.java" -prune -o -name "*.java" -print | ${SORT_DIRECTORY_ORDER} > javadoc-files.txt
 
 # If Java 8, don't run javadoc; a bug in javadoc causes it to crash.
 ifeq ($(JAVA8), 1)
   SKIP_JAVADOC=1
 endif
 
+# When JDK 24 is offically released, remove '--enable-preview'
+ifeq ($(shell test ${JAVA_RELEASE_NUMBER} -gt 23; echo $$?),0)
```
mernst, line 1290: Can this use `${extra-error-prone-args-2}` (which in that case should be renamed)?  Or does javadoc not accept `-source 24`?

Makefile
```diff
@@ -14,6 +14,12 @@ ifneq "$(wildcard ${SORT_DIRECTORY_ORDER})" "${SORT_DIRECTORY_ORDER}"
   SORT_DIRECTORY_ORDER = sort
 endif
 
+JAVA_VERSION_STRING := $(shell javac -version 2>&1 | head -1 | cut "-d " -f2 | sed 's/-ea//')
```
mernst, line 17: I think
```
java -version 2>&1 | head -1 | cut -d'"' -f2 | sed '/^1\./s///' | cut -d'.' -f1 | sed 's/-ea//'
```
works as a one-liner.  Does it not?

java/daikon/chicory/ChicoryTest.java
```diff
@@ -21,10 +23,17 @@ public static void main(String[] args) {
     double d = t.sample2();
     out.println("sample return [62.4] = " + d);
 
+    List<String> list = Arrays.asList(args);
+    for (String s : list) {
+      out.println("arg: " + s);
```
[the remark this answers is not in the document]
markro49: I'm not sure what you mean.  Comment out the println?

java/daikon/chicory/Instrument.java
```diff
@@ -596,7 +624,12 @@ private ClassInfo instrument_all_methods(ClassGen cg, String fullClassName, Clas
     return class_info;
   }
 
-  // This method exists only to suppress interning warnings
+  /**
+   * This method exists only to suppress interning warnings.
```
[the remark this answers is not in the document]
markro49: Method was added in 2011.  In 2019, Suzanne removed the @SupressWarning but left the method.  I think it can be deleted.

java/daikon/chicory/ChicoryTest.java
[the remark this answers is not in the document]
mernst: I wasn't sure what ChicoryTest.java did.  You can retain it.

java/daikon/chicory/Instrument.java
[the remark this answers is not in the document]
mernst: OK.  Please do so, in this pull request or in a different one.

java/daikon/chicory/MethodGen24.java
```diff
@@ -0,0 +1,347 @@
...
+
+  /**
+   * Constructor for the MethodGen24 object.
+   *
+   * @param methodModel for the method
+   * @param className for the containing class
+   * @param inst_obj the daikon.chicory.Instrument instance
+   */
+  public MethodGen24(
+      final MethodModel methodModel, final String className, daikon.chicory.Instrument24 inst_obj) {
```
mernst, line 93: I personally omit `final` because it is an implementation detail of the method body (it has nothing to do with the method's specification, which is what the signature should represent) and it clutters the signature.  But I don't object to you keeping it if you feel it is helpful.

java/daikon/chicory/MethodGen24.java
```diff
@@ -0,0 +1,347 @@
...
+
+    accessFlags = methodModel.flags();
```
mernst, line 95: Given that `methodModel` is stored, what is the point of storing its fields?  Does this reduce re-computation?

java/daikon/chicory/MethodGen24.java
```diff
@@ -0,0 +1,347 @@
...
+    // System.out.println("methodName: " + methodName);
+    descriptor = methodModel.methodType().stringValue();
+    this.className = className;
+    isStatic = accessFlags.has(AccessFlag.STATIC);
+
+    Optional<CodeModel> code = methodModel.code();
+    if (code.isPresent()) {
+      this.code = code.get();
+      // the original elementList is immutable, so we need to make a copy
+      this.codeList = new LinkedList<CodeElement>(this.code.elementList());
```
mernst, line 105: Why is this a LinkedList rather than an ArrayList?  A comment would be helpful.

java/daikon/chicory/MethodGen24.java
```diff
@@ -0,0 +1,347 @@
...
+
+    accessFlags = methodModel.flags();
+    methodName = methodModel.methodName().stringValue();
+    // System.out.println("methodName: " + methodName);
+    descriptor = methodModel.methodType().stringValue();
+    this.className = className;
+    isStatic = accessFlags.has(AccessFlag.STATIC);
+
+    Optional<CodeModel> code = methodModel.code();
+    if (code.isPresent()) {
```
mernst, line 102: This code style differs from that below.

java/daikon/chicory/MethodGen24.java
```diff
@@ -0,0 +1,347 @@
...
+      this.code = code.get();
+      // the original elementList is immutable, so we need to make a copy
+      this.codeList = new LinkedList<CodeElement>(this.code.elementList());
+    } else {
+      this.code = null;
+      this.codeList = new ArrayList<>();
+    }
+
+    codeAttribute = methodModel.findAttribute(Attributes.code()).orElse(null);
+    if (codeAttribute != null) {
```
mernst, line 112: I don't object to this style, or that above.  I do notice that they are different, though, and I think it would be better to be consistent.

java/daikon/chicory/MethodGen24.java
```diff
@@ -0,0 +1,347 @@
...
+    SignatureAttribute sa = methodModel.findAttribute(Attributes.signature()).orElse(null);
+    if (sa != null) {
+      signature = sa.signature().stringValue();
+    } else {
+      // if no signature then probably no type arguments so descriptor will do
+      signature = descriptor;
+    }
+
+    mtd = methodModel.methodTypeSymbol();
+    argTypes = mtd.parameterArray();
```
mernst: The use of both "arg" and "parameter" here indicates that one of the two names is not applicable.

java/daikon/chicory/MethodGen24.java
```diff
@@ -0,0 +1,345 @@
...
+   * Return the name of the containing class.
+   *
+   * @return class that contains this method
+   */
+  public String getClassName() {
+    return className;
+  }
+
+  /**
+   * Return the CodeAttribute for the method.
```
mernst, line 259: Please be more descriptive than the type.

java/daikon/chicory/MethodGen24.java
```diff
@@ -0,0 +1,345 @@
...
+   * Return the maximum stack size.
+   *
+   * @return the maximum stack size
+   */
+  public int getMaxStack() {
+    return maxStack;
+  }
+
+  /**
+   * Return the descriptor for the current method.
```
mernst, line 286: In what format?

java/daikon/chicory/MethodGen24.java
```diff
@@ -0,0 +1,345 @@
...
+   *
+   * @return descriptor for the current method
+   */
+  public String getDescriptor() {
+    return descriptor;
+  }
+
+  /**
+   * Return the signature for the current method.
```
mernst, line 295: In what format?

java/daikon/chicory/MethodGen24.java
```diff
@@ -0,0 +1,345 @@
...
+   * @return signature for the current method
+   */
+  public String getSignature() {
+    return signature;
+  }
+
+  /**
+   * Return the name of the containing class.
+   *
+   * @return class that contains this method
```
mernst, line 306: This documentation is wrong.

mernst (changes requested):
Here are some comments on `MethodGen24.java`.  I'll take another look after you have addressed these (and applied any that are applicable to `Instrument24.java`.
FYI, I have made renamings that I expect will cause compilation to fail; you will need to fix the compilation failures.
Thanks.

java/daikon/chicory/Instrument24.java
```diff
@@ -0,0 +1,1651 @@
...
+
+  /**
+   * Format a field descriptor for output. The main difference between a descriptor and a signature
+   * is that the later may contain type arguments. This routine was orginaly written for
+   * descriptors, but some support for type arguments has been added.
+   *
+   * @param descriptor the object to format
+   * @return a formatted string
+   */
+  public static String convertDescriptorToString(String descriptor) {
```
mernst, line 1464: Please rename this from "...ToString" to indicate what the output format is.  Also, describe that in the Javadoc.

java/daikon/chicory/Instrument24.java
```diff
@@ -0,0 +1,1651 @@
...
+      result.append(descriptor.substring(1, endOfBaseType).replace('/', '.'));
+    } else {
+      throw new IllegalArgumentException("Malformed object type descriptor: " + descriptor);
+    }
+
+    return result.toString();
+  }
+
+  /**
+   * Format a generic parameter for output.
```
mernst, line 1547: Should this be "Format one or more type parameters for output."?

java/daikon/chicory/Instrument24.java
```diff
@@ -0,0 +1,1651 @@
...
+    // Append array brackets if applicable
+    while (arrayDimensions-- > 0) {
+      result.append("[]");
+    }
+
+    return result.toString();
+  }
+
+  /**
+   * Format an object that may contain generics for output.
```
mernst, line 1518: What is the object?  Is it a type?

java/daikon/chicory/Instrument24.java
```diff
@@ -0,0 +1,1651 @@
...
+   *
+   * @param descriptor the object to format
+   * @return a formatted string
```
mernst, line 1521: In what format?

mernst (changes requested):
Thanks for this significant rewrite.  I have some suggestions and questions.

java/daikon/chicory/MethodGen24.java
```diff
@@ -0,0 +1,395 @@
...
```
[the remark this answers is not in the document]
mernst: The documentation is still wrong. It was cut-and-pasted from a method that returns a class name, but this method returns an instruction list.

mernst (changes requested):
I suspect that you haven't pushed your latest changes, because when reviewing this I re-noticed numerous issues that I had mentioned in previous pull requests.  Can you double-check that your changes are pushed?  Thanks.

java/daikon/chicory/Instrument24.java
```diff
@@ -0,0 +1,1720 @@
...
+  /** Directory into which to dump debug-instrumented classes. */
+  File debug_bin_dir;
+
+  /** Directory into which to dump original classes. */
+  File debug_orig_dir;
+
+  /** The index of this method in SharedData.methods. */
+  int cur_method_info_index = 0;
+
+  /** The MethodInfo for the current method. */
```
mernst, line 124: Under what circumstances is it null?

java/daikon/chicory/Instrument24.java
```diff
@@ -0,0 +1,1720 @@
...
+  @Nullable MethodInfo curMethodInfo;
```
mernst, line 125: I don't think this should be a field.  The value can be passed from `instrument_all_methods` to `instrumentCode`.

java/daikon/chicory/Instrument24.java
```diff
@@ -0,0 +1,1720 @@
...
+import org.checkerframework.checker.signature.qual.FieldDescriptor;
+import org.checkerframework.checker.signature.qual.FqBinaryName;
+import org.checkerframework.checker.signature.qual.InternalForm;
+import org.checkerframework.dataflow.qual.Pure;
+
+/**
+ * The Instrument24 class is responsible for modifying another class's bytecodes. Specifically, its
+ * main task is to add calls into the Chicory Runtime at method entries and exits for
+ * instrumentation purposes. These added calls are sometimes referred to as "hooks".
+ *
```
mernst, line 91: Please say something about how this class is used.  Is it instantiated anew for each class (so that each instance is for instrumenting one particular class), or does it have a method that takes a class to be instrumented?

java/daikon/chicory/Instrument24.java
```diff
@@ -0,0 +1,1651 @@
...
+  /** Current class name in binary format. */
+  @BinaryName String binaryClassName;
+
+  /** True if the class has a class initializer. */
+  private boolean hasClinit;
+
+  /** The current ClassModel. */
+  private ClassModel classModel;
+
+  /** Stores useful information about a class. */
```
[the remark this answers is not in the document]
mernst: In particular, about the current class?

java/daikon/chicory/Instrument24.java
```diff
@@ -0,0 +1,1651 @@
...
+  // Variables used for the entire class.
+
+  /** Current class name in binary format. */
+  @BinaryName String binaryClassName;
+
+  /** True if the class has a class initializer. */
+  private boolean hasClinit;
+
+  /** The current ClassModel. */
+  private ClassModel classModel;
```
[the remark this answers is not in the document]
mernst: This field is set in `transform`, but it is never used elsewhere. I suggest making it a local variable in that routine, to clarify data flow.

java/daikon/chicory/Instrument24.java
```diff
@@ -0,0 +1,1720 @@
...
+
+  /** Stores useful information about a class. */
+  private ClassInfo classInfo;
+
+  /** ConstantPool builder for entire class. */
+  private ConstantPoolBuilder poolBuilder;
+
+  // Variables used for the entire method.
```
mernst, line 159: Which method is "the method"?  Is it "the current method"?

java/daikon/chicory/Instrument24.java
```diff
@@ -0,0 +1,1651 @@
...
+  // They are initialized in modifyCode().
+
+  /** CodeBuilder for the current method. */
+  private CodeBuilder currentCodeBuilder;
```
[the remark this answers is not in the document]
mernst: This variable doesn't need to be a field. It can be passed from `modifyCode()` to `add_entry_instrumentation()`, which is the only place it is used. That will clarify the data flow and reduce the number of invariants about which methods may be called in which order.

Can you make similar changes to eliminate the other method-specific fields?
If there would be too many arguments passed to helper functions, then you could create a record that contains all these fields and pass it to the helper functions.

The same thing should be done for all the fields that hold information about the current class.

java/daikon/chicory/Instrument24.java
```diff
@@ -0,0 +1,1720 @@
...
+  /** Label for first byte code of method, used to give new locals method scope. */
+  private Label startLabel;
+
+  /** Label for last byte code of method, used to give new locals method scope. */
+  private Label endLabel;
+
+  /** Label for start of orignal code, post insertion of entry instrumentation. */
+  private Label entryLabel;
+
+  /** Label for first byte code of method, as a CodeModel Label. */
```
mernst, line 183: I presume this is for the original code of the method.

java/daikon/chicory/Instrument24.java
```diff
@@ -0,0 +1,1720 @@
...
+    }
+  }
+
+  /**
+   * Given a class, return a transformed version of the class that contains instrumentation code.
+   * Because Chicory is invoked as a javaagent, the transform method is called by the Java runtime
+   * each time a new class is loaded. A return value of null leaves the byte codes unchanged.
+   */
+  @Override
+  public byte @Nullable [] transform(
```
mernst, line 267: Can this be a static method?  How does it depend on instance state of the `Instrument24`?

java/daikon/chicory/Instrument24.java
```diff
@@ -0,0 +1,1720 @@
...
+      @Nullable ClassLoader loader,
+      @InternalForm String className,
+      @Nullable Class<?> classBeingRedefined,
+      ProtectionDomain protectionDomain,
+      byte[] classfileBuffer)
+      throws IllegalClassFormatException {
+
+    binaryClassName = Signatures.internalFormToBinaryName(className);
+
```
mernst, line 276: This variable doesn't need to be a class field.  It can be passed into the routines that use it, such as `call_initNotify` and `instrument_all_methods`.

java/daikon/chicory/Instrument24.java
```diff
@@ -0,0 +1,1720 @@
...
+      return null;
+    } else if (binaryClassName.startsWith("sun.reflect")) {
+      debug_transform.log("Ignoring system class %s, in sun.reflect package%n", binaryClassName);
+      return null;
+    } else if (binaryClassName.startsWith("jdk.internal.reflect")) {
+      // Starting with Java 9 sun.reflect => jdk.internal.reflect.
+      debug_transform.log(
+          "Ignoring system class %s, in jdk.internal.reflect package", binaryClassName);
+      return null;
+    }
```
mernst, line 310: I suggest abstracting this out into a `isBootClass()` method.

java/daikon/chicory/Instrument24.java
```diff
@@ -0,0 +1,1651 @@
...
+        try (BufferedWriter writer =
+            Files.newBufferedWriter(
+                new File(debug_orig_dir, binaryClassName + ".javap").toPath(),
+                StandardCharsets.UTF_8)) {
+          writer.write(classModel.toDebugString());
+        }
+      } catch (Throwable t) {
+        System.err.printf(
+            "Unexpected error %s dumping out debug files for: %s%n", t, binaryClassName);
+        t.printStackTrace();
```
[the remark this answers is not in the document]
mernst: It is inconsistent that the error message is printed to stderr, but the stack trace is printed to stdout. Make them consistent.

The same comment applies in other parts of the code.

java/daikon/chicory/Instrument24.java
```diff
@@ -0,0 +1,1720 @@
...
+              classModel.thisClass().asSymbol(),
+              classBuilder -> instrumentClass(classBuilder, classModel, cfLoader));
+
+    } catch (Throwable t) {
+      System.err.printf("Unexpected error %s in transform of %s%n", t, binaryClassName);
+      t.printStackTrace();
+      throw new RuntimeException("Unexpected error", t);
+    }
+
+    if (classInfo.shouldInclude) {
```
mernst, line 382: Can this value be the return value of `instrumentClass` rather than being stored in a field?

java/daikon/chicory/Instrument24.java
```diff
@@ -0,0 +1,1720 @@
...
+      return newBytes;
+    } else {
+      debug_transform.log("Didn't instrument %s%n", binaryClassName);
+      // No changes to the bytecodes
+      return null;
+    }
+  }
+
+  /**
+   * Instrument the current class.
```
mernst, line 415: Should this be "the given class"?  What is "the current class"?

java/daikon/chicory/Instrument24.java
```diff
@@ -0,0 +1,1720 @@
...
+        for (CodeElement ce : new_il) {
+          li.add(ce);
+        }
+
+        // skip over 'inst' we just inserted new_il in front of
+        li.next();
+      }
+    } catch (Exception e) {
+      System.err.printf("Unexpected exception encountered: %s", e);
+      e.printStackTrace();
```
mernst, line 497: Again, the first line prints to stderr and the second to stdout.  Is that intentional?  If so, what is the purpose?

java/daikon/chicory/Instrument24.java
```diff
@@ -0,0 +1,1720 @@
...
+  /**
+   * Called by {@link #addInvokeToClinit} to obtain the instructions that represent a call to the
+   * Chicory Runtime {@code initNotify} method prior to a return opcode. Returns null if the given
+   * instruction is not a return.
+   *
+   * @param inst the instruction that might be a return
+   * @return the list of instructions that call {@code initNotify}, or an empty list if {@code inst}
+   *     is not a return instruction
+   */
+  private List<CodeElement> xform_clinit(CodeElement inst) {
```
mernst, line 510: This is a short method, and it is only called once.  I suggest inlining it.

mernst (changes requested):

java/daikon/chicory/Instrument24.java
markro49: The ClassFileTransformer interface requires transform to be an instance method.  There is only a single instance of Instrument24 for the entire execution of Chicory.

java/daikon/chicory/Instrument24.java
markro49: instrument_all_methods=>addInvokeToClinit=>xform_clinit=>call_initnotify

seems like a lot of trouble just to eliminate a class field

java/daikon/chicory/Instrument24.java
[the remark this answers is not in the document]
markro49: Throwable.printStackTrace() goes to stderr.

java/daikon/chicory/Instrument24.java
markro49: I don't see the difference.  All of Instrument operates on a single class at a time via the ClassfileTransformer interface.

java/daikon/chicory/Instrument24.java
markro49: as noted before - default is stderr

java/daikon/chicory/MethodGen24.java
markro49: I think this comment is  out of date.

java/daikon/chicory/Instrument24.java
markro49: I don't see how to rewrite the 'newBytes = ..." to do that.

java/daikon/chicory/Instrument24.java
markro49: i have made binaryClassName local to transform

java/daikon/chicory/Instrument24.java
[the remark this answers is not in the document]
markro49: currentCodeBuilder is used in six different methods.  I would prefer to leave it as a class field.  I have removed two other class field items.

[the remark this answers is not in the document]
markro49: Sorry - I was looking at poolBuilder not codeBuilder - will continue to review.

markro49:
I think I have address all your comments.

mernst (commented):
I have made some changes.  Could you please look over them?  For example, should the `MInfo24` be a field in `MethodGen24`?

java/daikon/chicory/Instrument24.java
```diff
@@ -0,0 +1,1757 @@
...
+    }
+    MethodRefEntry mre = mgen.getPoolBuilder().methodRefEntry(runtimeCD, methodToCall, methodArgs);
+    newCode.add(InvokeInstruction.of(Opcode.INVOKESTATIC, mre));
+  }
+
+  /** Possibly modified default switch target. */
+  private Label modifiedTarget;
+
+  /** Possibly modified switch case list. */
+  private List<SwitchCase> modifiedCaseList;
```
mernst, line 1173: These two fields can be local variables.  `retargetStartLabel(Label defaultTarget, List<SwitchCase> caseList, MInfo24 minfo)` can return (instead of a boolean) either null or a struct containing these two fields.

java/daikon/chicory/Instrument24.java
```diff
@@ -0,0 +1,1757 @@
...
+
+  /**
+   * Insert the our instrumentation code into the instruction list for the given method.
+   *
+   * @param instructions instruction list for method
+   * @param mgen describes the given method
+   * @param curMethodInfo provides additional information about the method
+   * @param minfo for the given method's code
+   */
+  private void insertInstrumentation(
```
mernst, line 779: What is the relationship between `insertInstrumentation` and `instrumentCode`?  Why are their names so different, when they seem to have very similar specifications?

java/daikon/chicory/Instrument24.java
```diff
@@ -0,0 +1,1757 @@
...
+        SharedData.all_classes.add(classInfo);
+      }
+    } else { // not included
+      debug_transform.log("Trace info not added to class %s%n", classInfo);
+    }
+    classInfo.shouldInclude = shouldInclude;
+  }
+
+  /**
+   * Output the given method with no changes.
```
mernst, line 693: Output to where?  
What is the distinction between `mm` and `mgen`?

Same questions for `copyMethod` and `copyCode`.

mernst (changes requested):
Thanks, this looks good!

mernst (approved):
Thanks!

PR #601, opened 2025-03-03
Avoid resource leak

mernst:
@Calvin-L does this fix look right to you?

### DIFF
--- a/java/daikon/FileIO.java
+++ b/java/daikon/FileIO.java
@@ -1180,12 +1180,20 @@ public final class FileIO {
       try {
         return chicSocket.getInputStream();
       } catch (IOException e) {
-        chicSocket.close();
+        try {
+          chicSocket.close();
+        } catch (IOException closeException) {
+          // do nothing
+        }
         throw new RuntimeException("Unable to get Chicory's input stream", e);
       }
     } catch (IOException e) {
       if (chicSocket != null) {
-        chicSocket.close();
+        try {
+          chicSocket.close();
+        } catch (IOException closeException) {
+          // do nothing
+        }
       }
       throw new RuntimeException("Unable to create server", e);
     }
--- a/java/daikon/FileIO.java
+++ b/java/daikon/FileIO.java
@@ -1157,6 +1157,8 @@ public final class FileIO {
    */
   private static @Owning InputStream connectToChicory() {
 
+    Socket chicSocket = null;
+
     // bind to any free port
     try (ServerSocket daikonServer = new ServerSocket(0)) {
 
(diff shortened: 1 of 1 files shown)


java/daikon/FileIO.java
```diff
@@ -1179,9 +1180,21 @@ private static void warn_if_hierarchy_mismatch(PptMap all_ppts) {
...
         return chicSocket.getInputStream();
       } catch (IOException e) {
+        try {
+          chicSocket.close();
+        } catch (IOException closeException) {
+          // do nothing
+        }
         throw new RuntimeException("Unable to get Chicory's input stream", e);
       }
     } catch (IOException e) {
```
Calvin-L, line 1190: You may need to broaden this to `Exception`. With this change I see
```
../daikon/java/daikon/FileIO.java:1175: error: [required.method.not.called] @MustCall method close may not have been invoked on chicSocket or any of its aliases.
        chicSocket = daikonServer.accept();
                   ^
  The type of object is: java.net.Socket.
  Reason for going out of scope: possible exceptional exit due to daikonServer.close() with exception type java.lang.Exception
```

mernst: Thanks!  I made this change.

Calvin-L (approved):

java/daikon/FileIO.java
Calvin-L: @mernst I think the one on this line (1190) needs to change. It looks like you only changed the one on line 1194 below.

mernst:
@Calvin-L How does this look now?

Calvin-L:
I can confirm this closes the resource leak. :)

To keep the exception handling hairball as simple as possible, the affected method could simply re-throw IOException. (The one call to `connectToChicory()` seems equipped to handle IOException.) That would also avoid use of `Error`, which looks a little strange here.

Although, you would lose the precise exception messages, so it might lead to a worse user experience.

<details>

<summary>Simpler code without as many exception handlers</summary>

```java
  private static @Owning InputStream connectToChicory() throws IOException {
    Socket chicSocket = null;

    // bind to any free port
    try (ServerSocket daikonServer = new ServerSocket(0)) {
      // tell Chicory what port we have!
      System.out.println("DaikonChicoryOnlinePort=" + daikonServer.getLocalPort());
      daikonServer.setReceiveBufferSize(64000);
      daikonServer.setSoTimeout(5000);
      chicSocket = daikonServer.accept();
      return chicSocket.getInputStream();
    } catch (Exception e) {
      if (chicSocket != null) {
        try {
          chicSocket.close();
        } catch (Exception closeException) {
          // do nothing
        }
      }
      throw e;
    }
  }
```

</details>

mernst:
I wanted to retain the more specific error messages, but I have adopted your code (thanks!) and we will see whether it is a problem for users in practice.

PR #635, opened 2025-07-15
Run the `mypy` type-checker

mernst:
Merge after #634.


mernst:
> a little confusing as most of the diffs are same as #634

The [description](https://github.com/codespecs/daikon/pull/635#issue-3233808754) says to merge #634 first.  Once that is done, the diffs will correspond to the pull request title:  just type-checking.

PR #637, opened 2025-07-20
Add a test for Java calling native code via JNI

markro49:


mernst (approved):
Very nice PR.  Thanks!

PR #654, opened 2025-09-17
Add diagnostics

mernst:
## Summary by CodeRabbit

- Bug Fixes
  - Prevents a null-related crash when formatting object strings.
  - Ensures tracking decision logic returns the same result while adding per-point decision logging.

- Chores
  - Adds extensive debug/logging for class names, implemented interfaces, methods, and parameter details to aid troubleshooting.
  - Improves error messages when loading the instrumenter by including the instrumenter’s class name.


mernst:
The pull request is failing due to diagnostics accessing not-fully-initialized objects.  Maybe other changes need to be merged before this pull request, or this one needs more guards.

PR #685, opened 2025-10-03
Add additional version of DynComp for Java 24

markro49:


java/Makefile
```diff
@@ -754,6 +757,10 @@ very-clean: clean clean-generated-files
 	-rm -rf api
 	-rm -rf daikon/dcomp/std_dcomp_out.diff daikon/dcomp/std_dcomp_out.txt
 	-rm -rf daikon/chicory/ChicoryTest.inv.out
+	-rm -rf daikon/dcomp/debug/
+	-rm -rf daikon/dcomp/DcompTest.log
+	-rm -rf daikon/chicory/debug
+
```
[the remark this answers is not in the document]
markro49: The duplication does not exist.

java/daikon/dcomp/ClassGen24.java
```diff
@@ -0,0 +1,218 @@
...
+   * @param name the interface name (binary format)
+   */
+  public void addInterface(@BinaryName String name) {
+    ClassDesc ue = ClassDesc.of(name);
+    ClassEntry ce = classBuilder.constantPool().classEntry(ue);
+    if (interfaceList.contains(ce)) {
+      return;
+    }
+    interfaceList.add(ce);
+  }
```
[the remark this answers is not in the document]
markro49: removed the static and added getInterfaceList method

java/daikon/chicory/ClassInfo.java
```diff
@@ -25,6 +25,9 @@ public class ClassInfo {
   /** True if the class has a class initializer. */
   public boolean hasClinit;
 
+  /** True if the class is a JUnit test class. */
+  public boolean isJunitTestClass;
```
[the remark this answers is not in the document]
markro49: fixed

java/daikon/dcomp/ClassGen24.java
```diff
@@ -0,0 +1,225 @@
...
+ * (no more fiddling with StackMaps) and are always up to date with any .class file changes (since
+ * they are part of the JDK). (We will need to continue to support Instrument.java using BCEL, as we
+ * anticipate our clients using JDK 21 or less for quite some time.)
+ */
+public class ClassGen24 {
+
+  /**
+   * Models the body of the class.
+   *
+   * <p>Several fields of ClassModel are declared as fields of ClassGen24 to better correspond to
```
mernst: Please put a `//` code comment before the first of those fields and after the last of those fields.

java/daikon/chicory/MethodGen24.java
mernst, line 94: Please put a `//` code comment before the first of those fields and after the last of those fields.

mernst: Please put a `//` code comment before the first of those fields and after the last of those fields.

mernst: Is it just this field, or multiple fields?  Please clarify, and maybe add a comment at the end of the fields so readers know how many there are.

java/daikon/chicory/MethodGen24.java
```diff
@@ -358,6 +620,15 @@ public ClassDesc[] getParameterTypes() {
     return paramTypes.clone();
```
mernst, line 641: I don't see anywhere that this return value is modified.  Is the use of `clone()` needed?  The same question goes for a number of other uses of `clone()`.

java/daikon/chicory/MethodGen24.java
mernst: It has "information about".  Does it contain the bytecodes themselves?  If so, you can increase brevity by dropping "information about", here and elsewhere.

mernst (changes requested):

java/daikon/chicory/MethodGen24.java
markro49: CodeRabbit asked for the '.clone()'.  But it worked fine without it which suggests you are correct.  Would you like it removed?

mernst: `clone()` is good defensive programming.  It is essential if the spec is "clients may do whatever they want".  If the spec is "clients should treat this as read-only", then it is merely defensive programming, and it may be more expense than we wish to pay.  (Ideally, we would have a verifier to ensure that the client didn't modify it, but we do not have that.)

markro49: Ok, let's leave it in.

java/daikon/dcomp/DCInstrument.java
```diff
@@ -4203,7 +4190,7 @@ MethodGen create_dcomp_stub(MethodGen mg) {
    */
   static void save_static_field_id(File file) throws IOException {
 
-    PrintStream ps = new PrintStream(file);
+    PrintStream ps = new PrintStream(file, "UTF-8");
```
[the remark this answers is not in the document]
markro49: Not supported in JDK 8.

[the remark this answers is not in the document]
mernst: @markro Thanks.

java/daikon/dcomp/BcelUtils24.java
```diff
@@ -0,0 +1,227 @@
...
+import java.util.List;
+import java.util.ListIterator;
+import org.checkerframework.checker.signature.qual.Identifier;
+
+/**
+ * This class provides utility methods for manipulating bytecode structures in the Java 24
+ * instrumentation pipeline, including operations on local variables, parameter types, and
+ * instruction adjustments.
+ */
+public final class BcelUtils24 {
```
mernst: Document the reason for this name and/or the relationship to the `BcelUtil.java` (singular, not plural, name) file in bcel-util.

java/daikon/dcomp/BcelUtils24.java
```diff
@@ -0,0 +1,227 @@
...
+  /**
+   * Process the instruction list, adding size (1 or 2) to the slot number of each Instruction that
+   * references a local that is equal or higher in the local map than offsetFirstMovedlocal. Size
+   * should be the size of the new local that was just inserted at offsetFirstMovedlocal.
+   *
+   * @param mgen MethodGen to be modified
+   * @param offsetFirstMovedLocal original offset of first local moved "up"
+   * @param size size of new local added (1 or 2)
+   */
+  static void adjust_code_for_locals_change(MethodGen24 mgen, int offsetFirstMovedLocal, int size) {
```
mernst: In bcel-util, this method is defined in StackMapUtils.java rather than in BcelUtil.java.  Therefore, should it be defined in a file StackMapUtils24.java here (if you are trying to keep this in sync with bcel-util; I'm not sure whether you are)?

java/daikon/dcomp/BcelUtils24.java
```diff
@@ -0,0 +1,227 @@
...
+    return newArr;
+  }
+
+  /**
+   * Process the instruction list, adding size (1 or 2) to the slot number of each Instruction that
+   * references a local that is equal or higher in the local map than offsetFirstMovedlocal. Size
+   * should be the size of the new local that was just inserted at offsetFirstMovedlocal.
+   *
+   * @param mgen MethodGen to be modified
+   * @param offsetFirstMovedLocal original offset of first local moved "up"
```
mernst: What is the meaning of "First" in this variable name?  Should you rename to `offsetOfMovedLocal`?

java/daikon/dcomp/BcelUtils24.java
```diff
@@ -0,0 +1,227 @@
...
+      if (debugInstrument.enabled) {
+        t.printStackTrace();
+      }
+      throw new DynCompError(
+          "Unexpected exception encountered in adjust_code_for_locals_change", t);
+    }
+  }
+
+  /**
+   * Add a new local variable to the method. This will be inserted after last current parameter and
```
mernst: Is "last current" a typo?

java/daikon/dcomp/BcelUtils24.java
```diff
@@ -0,0 +1,227 @@
...
+   * before the first local variable. This might have the side effect of causing us to rewrite the
```
mernst: "might": doesn't it *always* do this?

java/daikon/dcomp/StackMapUtils24.java
```diff
@@ -0,0 +1,227 @@
...
+      newOffset++; // size of 'this' is 1
+    }
+
+    if (argTypes.length > 0) {
+      LocalVariable lastArg;
+      newIndex = newIndex + argTypes.length;
+      // newIndex is now positive, because argTypes.length is
+      lastArg = locals.get(newIndex - 1);
+      newOffset = lastArg.slot() + TypeKind.from(lastArg.typeSymbol()).slotSize();
+    }
```
mernst, line 148: Abstract out into a method that computes the offset after the last formal parameter.

java/daikon/dcomp/StackMapUtils24.java
```diff
@@ -0,0 +1,227 @@
...
+   * @return a LocalVariable for the new variable
+   */
+  public static LocalVariable addNewSpecialLocal(
+      MethodGen24 mgen,
+      String argName,
+      ClassDesc argType,
+      MethodGen24.MInfo24 minfo,
+      boolean isParam) {
+
+    debugInstrument.enabled = daikon.dcomp.DCInstrument24.bcelDebug;
```
mernst: I think this belongs in the caller, not here in a utility routine.  (This couples the utility routines too closely to `DCInstrument24.java`)

java/daikon/dcomp/StackMapUtils24.java
```diff
@@ -0,0 +1,227 @@
...
+      // Update the method's parameter information.
+      argTypes = postpendToArray(argTypes, argType);
+      @SuppressWarnings("signature:assignment") // string manipulation
+      @Identifier String[] argNames = addString(mgen.getParameterNames(), argName);
+      mgen.setParameterTypes(argTypes);
+      mgen.setParameterNames(argNames);
+    }
+
+    String varType = isParam ? "arg" : "local";
+    debugInstrument.log(
```
mernst: This combination of `%s` and `+` is surprising; I suggest using one or the other throughout the statement.

java/daikon/dcomp/CalcStack24.java
```diff
@@ -0,0 +1,1047 @@
...
+import java.lang.classfile.instruction.StoreInstruction;
+import java.lang.classfile.instruction.SwitchCase;
+import java.lang.classfile.instruction.TableSwitchInstruction;
+import java.lang.classfile.instruction.TypeCheckInstruction;
+import java.lang.constant.ClassDesc;
+import java.lang.constant.MethodTypeDesc;
+import java.util.List;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+/** This class calculates the state of the operand stack via simulation. */
```
mernst: The documentation should indicate that the methods of this class side-effect data structures in `DCInstrument24`.

java/daikon/dcomp/CalcStack24.java
```diff
@@ -0,0 +1,1047 @@
...
+   *
+   * @param mgen method containing the instruction (currently unused)
+   * @param minfo for the given method's code (currently unused)
+   * @param ce instruction to be interpreted
+   * @param inst_index index of ce in code element list
+   * @param stack current state of operand stack
+   * @return true if should continue with next instruction, false otherwise
+   */
+  @SuppressWarnings("fallthrough")
+  static boolean calcOperandStack(
```
mernst: A better name for this method would be "execute" or "interpret"

mernst: I think this method should be an instance method of `OperandStack24`.  Then, this file does not need to exist.

java/daikon/dcomp/CalcStack24.java
```diff
@@ -0,0 +1,1047 @@
...
+      CodeElement ce,
+      int inst_index,
+      OperandStack24 stack) {
+
+    if (DCInstrument24.debugOperandStack) {
+      System.out.println("code element: " + ce);
+    }
+    switch (ce) {
+      case Instruction inst -> {
+        if (DCInstrument24.stacks[inst_index] != null) {
```
mernst: To make the overall structure of this method clearer (that is, to show what happens for cases other than `case Instruction inst -> {`), I suggest abstracting out the body of `case Instruction inst -> {` into a method that takes an `Instruction`.

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+  /** Mapping from a label to its index in the method's codeList. */
+  protected static Map<Label, Integer> labelIndexMap;
+
+  /** Mapping from a label to its operand stack. */
+  protected static Map<Label, OperandStack24> worklistHistory;
+
+  /** State of operand stack prior to each byte code instruction. */
+  protected static OperandStack24[] stacks;
+
+  /** The type of each local variable. */
```
mernst: Each local variable of what?  What does `DCInstrument24.java` represent?  If the class represents something, should this field be an instance field?

java/daikon/dcomp/CalcStack24.java
```diff
@@ -0,0 +1,1047 @@
...
+        switch (inst.opcode()) {
+
+          // operand stack:
+          // ..., arrayref, index
+          // ..., value
+          case Opcode.AALOAD:
+            {
+              stack.pop(); // discard the index
+              final ClassDesc t = stack.pop(); // pop the arrayref
+              if (t == null || nullCD.equals(t)) {
```
mernst: Throughout, I think it's a bug if `t` is `null`, and an error should be thrown in that case.

java/daikon/dcomp/CalcStack24.java
```diff
@@ -0,0 +1,1047 @@
...
+                // Push nullCD to maintain stack consistency. The actual NullPointerException
```
mernst: What does it mean to maintain stack consistency?  I'm unsure why it is important to track whether a value is null.  Even if the value is statically known to be null, it is known to be null of some specific type.  So, why not always push `ct` here, even if the value is known to be null?

This comment may indicate that I am missing something important about the algorithm.

java/daikon/dcomp/CalcStack24.java
```diff
@@ -0,0 +1,1047 @@
...
+            stack.push(CD_double);
+            return true;
+
+          // operand stack:
+          // ..., value
+          // ..., value, value
+          case Opcode.DUP:
+            {
+              final ClassDesc t = stack.pop();
+              stack.push(t);
```
mernst: Could this just call `peek()`?  Or maybe you think the explicit `pop` then `push` is more informative.

java/daikon/dcomp/CalcStack24.java
```diff
@@ -0,0 +1,1047 @@
...
+          case Opcode.ILOAD_W:
+            stack.push(CD_int);
+            return true;
+
+          // operand stack:
+          // ..., value
+          // ..., result
+          case Opcode.INEG:
+            stack.pop(); // discard the value
+            stack.push(CD_int);
```
mernst: Would it work to leave the stack as is?  There must have been an int on the stack before the operation, so it seems that the types of the stack elements are unchanged by this operation.

Again, I might be missing something.  And I don't object to this code, in that it seems to operate correctly.  I'm wondering about design and corner cases.

java/daikon/dcomp/CalcStack24.java
```diff
@@ -0,0 +1,1047 @@
...
+              case ConstantDynamicEntry cde -> stack.push(cde.typeSymbol());
+              case DoubleEntry de ->
+                  stack.push(CD_double); // LDC2_W only, but we assume correct code
+              case FloatEntry fe -> stack.push(CD_float);
+              case IntegerEntry ie -> stack.push(CD_int);
+              case LongEntry le -> stack.push(CD_long); // LDC2_W only, but we assume correct code
+              case MethodHandleEntry mhe ->
+                  stack.push(ClassDesc.of("java.lang.invoke.MethodHandle"));
+              case MethodTypeEntry mte -> stack.push(ClassDesc.of("java.lang.invoke.MethodType"));
+              case StringEntry se -> stack.push(CD_String);
```
mernst: I think this would be clearer abstracted out as a method that takes a LoadableConstantEntry and returns a class descriptor.  Then the client here can just be: `stack.push(lceToCD(lce))`.

java/daikon/dcomp/CalcStack24.java
```diff
@@ -0,0 +1,1047 @@
...
+            return true;
+
+          // operand stack:
+          // ..., count
+          // ..., arrayref
+          case Opcode.NEWARRAY:
+            stack.pop(); // discard the count
+            final NewPrimitiveArrayInstruction npai = (NewPrimitiveArrayInstruction) inst;
+            String descriptor;
+            switch (npai.typeKind()) {
```
mernst: Can you abstract out a method that goes from a TypeKind to a descriptor?

java/daikon/dcomp/CalcStack24.java
```diff
@@ -0,0 +1,1047 @@
...
+            // the variable referenced must contain a return address
+            // that we would treat as target of jump; no way to do that
+            return false;
+
+          // operand stack:
+          // ..., value1, value2
+          // ..., value2, value1
+          case Opcode.SWAP:
+            {
+              final ClassDesc v2 = stack.pop();
```
mernst: Minor: `v1` and `v2` are actually types (well, ClassDesc) rather than values.  Thus, the use of variable names `v` is a bit misleading.  But, be consistent throughout the whole file.

java/daikon/dcomp/CalcStack24.java
```diff
@@ -0,0 +1,1047 @@
...
+      }
+
+      default -> {
+        throw new DynCompError("Unexpected CodeElement: " + ce);
+      }
+    }
+  }
+
+  /**
+   * Create a set of worklist items.
```
mernst: Where is it created?

java/daikon/dcomp/ClassGen24.java
```diff
@@ -0,0 +1,228 @@
...
+  // Start of ClassModel items.
+
+  /** The class's access flags. */
+  private AccessFlags accessFlags;
+
+  /** The list of interfaces this class implements. */
+  private List<ClassEntry> interfaceList;
+
+  /** The name of the class's enclosing class, in binary name format. */
+  private @BinaryName String superclassName;
```
mernst: The "enclosing class" is not the same as the superclass, so the documentation and variable name are inconsistent.  Below, the constructor calls it the "containing class".  Please be consistent throughout.

java/daikon/dcomp/OperandStack24.java
```diff
@@ -0,0 +1,233 @@
...
+ * <p>This class implements a stack used for symbolic JVM stack simulation. [It's used as an operand
+ * stack substitute.] Elements of this stack are {@link ClassDesc} objects.
+ */
+public class OperandStack24 implements Cloneable {
+
+  /** We hold the stack information here. */
+  private ArrayList<ClassDesc> stack = new ArrayList<>();
+
+  /** The maximum number of stack slots this OperandStack instance may hold. */
+  private final @NonNegative int maxStack;
```
mernst, line 43: I don't see how this is used.  That is, I don't see any enforcement anywhere.  Did I miss it in my scan of the code, or can this field be removed?

java/daikon/dcomp/OperandStack24.java
```diff
@@ -0,0 +1,233 @@
...
+
+  /** Creates an empty stack with a maximum of maxStack slots. */
+  public OperandStack24(final @NonNegative int maxStack) {
+    this.maxStack = maxStack;
+  }
+
+  /** Clears the stack. */
+  public void clear() {
+    stack = new ArrayList<>();
```
mernst: Why not use `stack.clear()`?

java/daikon/dcomp/OperandStack24.java
```diff
@@ -0,0 +1,233 @@
...
+      throw new DynCompError("Unexpected error: ", e);
+    }
+    newstack.stack = new ArrayList<>(this.stack);
+    return newstack;
+  }
+
+  /**
+   * Returns true if and only if this OperandStack equals another, meaning equal lengths and equal
+   * objects on the stacks. Special case 'null' which matches anything but a primitive.
+   */
```
mernst, line 81: In the documentation above, why is the special case needed?  Please explain, somewhere.

java/daikon/dcomp/OperandStack24.java
```diff
@@ -0,0 +1,233 @@
...
+        // We assume an array matches any array or any class as they all have Object as super class
+        // at some point.
+      } else if (thisItem.isClassOrInterface()) {
+        if (otherItem.isPrimitive()) {
+          return false;
+        }
+        // We assume a class matches any array or any class as they all have Object as super class
+        // at some point.
+      } else if (thisItem.isPrimitive() && !otherItem.isPrimitive()) {
+        return false;
```
mernst: Please abstract this out into a "compareElement" method in OperandStack24.java.

java/daikon/dcomp/OperandStack24.java
```diff
@@ -0,0 +1,233 @@
...
+        // We don't bother to check they are the same primitive.
+      }
+    }
```
mernst, line 100: Why don't we bother to check?  I think it's because we assume the given bytecode is correct and maintains the run-time stack correctly.  That doesn't mean that this class has maintained the operand stack correctly, though, and that is where the test could be useful.

java/daikon/dcomp/OperandStack24.java
```diff
@@ -0,0 +1,233 @@
...
+  }
+
+  /** Returns the size of this OperandStack; that means, how many ClassDesc objects there are. */
+  @Pure
+  public @NonNegative int size(@GuardSatisfied OperandStack24 this) {
+    return stack.size();
+  }
+
+  /**
+   * Returns the number of stack slots used.
```
mernst: Does this always return 1 or 2?

java/daikon/dcomp/OperandStack24.java
```diff
@@ -0,0 +1,233 @@
...
+      sb.append(peek(i)).append(" (Size: ").append(String.valueOf(slotSize(peek(i)))).append(")\n");
+    }
+    return sb.toString();
+  }
+
+  /**
+   * Calculate the size of an item on the operand stack.
+   *
+   * @param item type to get size of
+   * @return size of item
```
mernst, line 280: Is this always 1 or 2?

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+ * its main task is to add calls into the DynComp Runtime to calculate comparability values. These
+ * added calls are sometimes referred to as "hooks".
+ *
+ * <p>Instrument24 and DCInstrument24 use Java's ({@code java.lang.classfile}) APIs for reading and
+ * modifying .class files. Those APIs were added in JDK 24. Compared to BCEL, these APIs are more
+ * complete and robust (no more fiddling with StackMaps) and are always up to date with any .class
+ * file changes (since they are part of the JDK). (We will need to continue to support
+ * Instrument.java using BCEL, as we anticipate our clients using JDK 21 or less for quite some
+ * time.)
+ */
```
mernst, line 357: Please discuss the relationship between `DCInstrument24.java` and `Instrument24.java`, in the Javadoc of both classes.  Right now their class Javadoc is quite similar.  The same comment probably applies to `DCInstrument.java` and `Instrument.java`.

java/daikon/dcomp/Instrument24.java
```diff
@@ -0,0 +1,353 @@
...
+      return true;
+    }
+    return false;
+  }
+
+  /**
+   * Returns true if the specified class is part of a tool known to do Java byte code
+   * transformation. We need to warn the user that this may not work correctly.
+   *
+   * @param classname class to be checked
```
mernst: In what format?

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+  public static boolean bcelDebug;
+
+  /**
+   * Used when testing to continue processing if an error occurs. Currently, This flag is only used
+   * by BuildJDK.
+   */
+  @Option("Halt if an instrumentation error occurs")
+  public static boolean quit_if_error = true;
+
+  /** The index of this method in SharedData.methods. */
```
mernst: What is "this method"?  The Javadoc of this class suggests that it represents a class rather than a method.

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+  int cur_method_info_index = 0;
+
+  /** Start of user code for current method, prior to instrumenting, as a CodeModel Label. */
+  protected @MonotonicNonNull Label oldStartLabel;
+
+  /** Start of user code for current method, after instrumenting, as a CodeBuilder Label. */
```
mernst, line 378: Inconsistent with "a CodeModel Label" above.

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+   */
+  protected int newStartIndex;
+
+  // Variables used for calculating the state of the operand stack.
+
+  /** Mapping from a label to its index in the method's codeList. */
+  protected static Map<Label, Integer> labelIndexMap;
+
+  /** Mapping from a label to its operand stack. */
+  protected static Map<Label, OperandStack24> worklistHistory;
```
mernst: Why does this variable have this name?  How is it related to the `worklist` field below?

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+  /** State of operand stack prior to each byte code instruction. */
+  protected static OperandStack24[] stacks;
+
+  /** The type of each local variable. */
+  protected static ClassDesc[] locals;
+
+  /** Record containing a work item for the operand stack calculation. */
+  record WorkItem(int instructionIndex, OperandStack24 stack) {}
+
+  /** Comparator to sort WorkItems with lowest address first. */
```
mernst: "address" is inconsistent with "instructionIndex" immediately above.

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+  protected static Comparator<WorkItem> indexComparator =
+      new Comparator<>() {
+        @Override
+        public int compare(WorkItem w1, WorkItem w2) {
+          return Integer.compare(w1.instructionIndex, w2.instructionIndex);
+        }
+      };
+
+  /** Queue of items for the operand stack calculation. */
```
mernst: Is this per method?

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+  private boolean trackClass = false;
+
+  /** ClassGen for the current class. */
+  protected ClassGen24 classGen;
+
+  /** Is the current class a member of the JDK? */
+  protected boolean in_jdk;
+
+  /** Has an {@code <init>} method completed initialization? */
+  protected boolean constructor_is_initialized;
```
mernst, line 467: I wonder whether this belongs in a new abstraction "ExecutionState" which includes OperandStack.

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+
+  /** Record containing subset of LocalVariable fields. */
```
mernst: Which subset?  How is it used?

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+  /** Possible states of JUnit test discovery. */
+  protected enum JunitState {
+    /** Have not seen a JUnit class file. */
+    NOT_SEEN,
+    /** Have seen a JUnit class file. */
+    STARTING,
+    /** Have seen a JUnit class file that loads JUnit test classes. */
+    TEST_DISCOVERY,
+    /** Have completed identifying JUnit test classes. */
+    RUNNING
```
mernst: The name "running" conflicts with the documentation "have completed".

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+  protected static JunitState junit_state = JunitState.NOT_SEEN;
+
+  /** Have we seen 'JUnitCommandLineParseResult.parse'? */
+  protected static boolean junit_parse_seen = false;
+
+  /**
+   * Map from each static field name to its unique integer id. Note that while it's intuitive to
+   * think that each static should show up exactly once, that is not the case. A static defined in a
+   * superclass can be accessed through each of its subclasses. Tag accessor methods must be added
+   * in each subclass and each should return the same id. We thus will lookup the same name multiple
```
mernst: What method is "we" referring to?

mernst: "the same name":  Is the idea here that there are different names for the same class?

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+   * times.
+   */
+  static Map<String, Integer> static_field_id = new LinkedHashMap<>();
```
mernst: String in what format?

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+
+  /**
+   * Map from class name to its access_flags. Used to cache the results of the lookup done in {@link
+   * #getAccessFlags}. If a class is marked ACC_ANNOTATION then it will not have been instrumented.
+   */
+  static Map<String, Integer> accessFlags = new HashMap<>();
```
mernst: String in what format?

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+        return false;
+      }
+      if (this.arg_types.length != arg_types.length) {
+        return false;
+      }
+      for (int ii = 0; ii < arg_types.length; ii++) {
+        if (!arg_types[ii].equals(this.arg_types[ii])) {
+          return false;
+        }
+      }
```
mernst: Why doesn't this use `Arrays.equals()`?

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+      for (ClassDesc arg : arg_types) {
+        code += arg.hashCode();
+      }
+      return code;
+    }
+  }
+
+  /** Initialize with the original class and whether or not the class is part of the JDK. */
+  @SuppressWarnings({
+    "nullness:StaticAssignmentInConstructor", // instrumentation_interface
```
mernst: Could you avoid the warnings by using a static initializer?

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+    }
+    dcompRuntimeClassname = Signatures.addPackage(dcomp_prefix, "DCRuntime");
+    DCRuntime.instrumentation_interface = Signatures.addPackage(dcomp_prefix, "DCompInstrumented");
+
+    // Turn on some of the logging based on debug option.
+    debugInstrument.enabled = DynComp.debug || Premain.debug_dcinstrument;
+    debug_native.enabled = DynComp.debug;
+    debug_transform.enabled = daikon.dcomp.Instrument24.debug_transform.enabled;
+
+    // TEMPORARY
```
mernst: Should this be removed?

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+    // Don't know where I got this idea.  They are executed.  Don't remember why
+    // adding dcomp marker causes problems.
+    // Don't instrument annotations.  They aren't executed and adding
+    // the marker argument causes subtle errors
+    if (classModel.flags().has(AccessFlag.ANNOTATION)) {
+      debug_transform.log("Not instrumenting annotation %s%n", classInfo.class_name);
+      return null;
+    }
+
+    // If a class has an EvoSuite annotation it may be instrumented by Evosuite;
```
mernst, line 781: Does this belong here?  Aren't there tests elsewhere for classes that shouldn't be instrumented?  It seems the logic might belong there.

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+   * @param classModel for the class
+   * @param classInfo for the given class
+   */
+  private void instrumentClass(
+      ClassBuilder classBuilder, ClassModel classModel, ClassInfo classInfo) {
+
+    @BinaryName String classname = classInfo.class_name;
+    classGen = new ClassGen24(classModel, classname, classBuilder);
+    runtimeCD = ClassDesc.of(dcompRuntimeClassname);
+    poolBuilder = classBuilder.constantPool();
```
mernst, line 823: Setting class fields in a method is a code smell.  Is there some method-specific data structure that could be passed around?  Could a new DCInstrument24 be created for every class, so it can have fields that aren't set after the constructor returns?

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+      debugInstrument.log("  %s%n", ce.asInternalName());
+    }
+
+    // Handle object methods for this class
+    handle_object(classGen);
+
+    trackClass = false;
+    boolean junit_test_class = false;
+
+    if (!in_jdk) {
```
mernst, line 847: Abstract this logic into its own method.

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+    for (Attribute<?> a : classModel.attributes()) {
+      debugInstrument.log("  %s%n", a);
+    }
+
+    debugInstrument.log("Class Interfaces:%n");
+    for (ClassEntry ce : classModel.interfaces()) {
+      debugInstrument.log("  %s%n", ce.asInternalName());
+    }
+
+    // Handle object methods for this class
```
mernst: Are these methods called in the class?  Defined by the class?

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+    // Have all top-level classes implement our interface
+    if (classGen.getSuperclassName().equals("java.lang.Object")) {
+      @SuppressWarnings("signature:assignment")
+      @MethodDescriptor String objectArgReturnBoolean = "(Ljava/lang/Object;)Z";
+      // Add equals method if it doesn't already exist. This ensures
+      // that an instrumented version, equals(Object, DCompMarker),
+      // will be created in this class.
+      MethodModel eq = classGen.containsMethod("equals", objectArgReturnBoolean);
+      if (eq == null) {
+        debugInstrument.log("Added equals method%n");
```
mernst: It hasn't happened yet.  Change this to "Adding" or move it after the call.

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+    }
+
+    // Output the list of interfaces.
+    classBuilder.withInterfaces(classGen.getInterfaceList());
+
+    // Copy all other ClassElements to output class unchanged.
+    for (ClassElement ce : classModel) {
+      debugInstrument.log("ClassElement: %s%n", ce);
+      switch (ce) {
+        case MethodModel mm -> {}
```
mernst, line 1043: Above, there was a loop over all methods, adding instrumentation and adding them to `classBuilder`.  Why is that separate loop necessary?  Would it be possible to operate on each method here, one at a time?  Or is the bulk processing of them all together essential?

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+        }
+      }
+    }
+
+    debug_transform.exdent();
+    debug_transform.log("Instrumentation complete: %s%n", classInfo.class_name);
+  }
+
+  /**
+   * Adds a call to the DynComp Runtime {@code set_class_initialized} method at the begining of the
```
mernst: "set_class_initialized" is inconsistent with "call_initNotify".  Please use consistent terminology.

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+
+    @BinaryName String classname = classInfo.class_name;
+
+    if (classModel.majorVersion() < ClassFile.JAVA_6_VERSION) {
+      System.out.printf(
+          "DynComp warning: ClassFile: %s - classfile version (%d) is out of date and may not be"
+              + " processed correctly.%n",
+          classname, classModel.majorVersion());
+      // throw new DynCompError("Classfile out of date");
+    }
```
mernst: I think this belongs at the call site, not here in the calleee.

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+
+    // Process each method in the class
+    for (MethodModel mm : classModel.methods()) {
+
+      try {
```
mernst: Please abstract out a method that operates on just one method at a time, and call it within the loop.

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+          }
+          debugInstrument.log("mgen.getSignature: %s%n", mgen.getSignature());
+          MethodTypeDesc mtd = mm.methodTypeSymbol();
+          debugInstrument.log("mtd.descriptorString: %s%n", mtd.descriptorString());
+          debugInstrument.log("mtd.displayDescriptor: %s%n", mtd.displayDescriptor());
+        }
+
+        boolean track = false;
+        if (!in_jdk) {
+          // Note whether we want to track the daikon variables in this method
```
mernst: What are "the daikon variables"?

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+          track = should_track(classname, mgen.getName(), methodEntryName(classname, mgen));
+
+          // We do not want to track bridge methods the compiler has synthesized as
```
mernst: Above talked about tracking specific variables.  Here is a discussion of tracking methods.  Below is a comment about tracking classes.  Please define all these terms somewhere.

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+                codeBuilder -> copyCode(codeBuilder, mgen.getInstructionList(), null));
+
+        // copy all other MethodElements to output class (unchanged)
+        default -> methodBuilder.with(me);
+      }
+    }
+  }
+
+  /**
+   * Copy an instruction list into the given method.
```
mernst: Does this append to the instructions in the given method, or replace them?

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+    }
+
+    for (CodeElement ce : instructions) {
+      debugInstrument.log("CodeElement: %s%n", ce);
+      codeBuilder.with(ce);
+    }
+  }
+
+  /**
+   * Instrument the given method using {@link #instrumentCode}.
```
mernst: Is this comment saying that `instrumentCode` does the real work and this is a wrapper around it?  What is the relationship between the two methods?

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+  /**
+   * Generate instrumentation code for the given method. This includes reading in and processing the
+   * original instruction list, calling {@code insertInstrumentationCode} to add the instrumentation
+   * code, and then copying the modified instruction list to the output method while updating the
+   * code labels, if needed. If DCInstrument24 generated the method or if it is a native method,
+   * then codeModel == null. If DCInstrument24 generated the method then localsTable will be set.
+   *
+   * @param codeBuilder for the given method's code
+   * @param codeModel for the input method's code; may be null
+   * @param newLocals method scope locals to define; may be null if none
```
mernst, line 1498: The comment above uses "localsTable"; should it use `newLocals` instead?

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+  }
+
+  /**
+   * Returns true if the specified classname.method_name is the root of JUnit startup code.
+   *
+   * @param classname class to be checked
+   * @param method_name method to be checked
+   * @return true if the given method is a JUnit trigger
+   */
+  boolean isJunitTrigger(String classname, String method_name) {
```
mernst: Inconsistent naming conventions:  use of underscore vs. not.  (Using camelCase rather than an underscore is more idiomatic Java.)

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+
+  /**
+   * Instrument the specified method for dynamic comparability.
+   *
+   * @param codeModel for the method's code
+   * @param mgen describes the given method
+   * @param minfo for the given method's code
+   * @param instructions instruction list for method
+   */
+  private void instrumentCodeList(
```
mernst, line 1821: Would `instrumentMethod` be a better name?

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+      }
+      labelIndexMap.put(newStartLabel, newStartIndex);
+
+      // Create an array to hold the calculated operand stack
+      // prior to each byte code instruction.
+      stacks = new OperandStack24[instructions.size()];
+
+      // Create an array containing the type of each local variable.
+      // This will be indexed by the local variable's slot number.
+      // There may be a gap for the second slot of a variable of type long or double.
```
mernst: Does this mean that slot numbers are not sequential?  If so, they sound more like slot offsets than slot numbers or slot indexes.

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+  protected static void verifyOperandStackMatches(
+      Label target, OperandStack24 existing, OperandStack24 current) {
+    if (existing.equals(current)) {
+      if (debugOperandStack) {
+        System.out.println("operand stacks match at: " + target);
+      }
+      return;
+    }
+    // stacks don't match
+    if (debugOperandStack) {
```
mernst: I would think that if the stacks don't match, we always want to produce this output, whether or not debugging is enabled.

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+  void skip_method(String method) {
+    skipped_methods.add(method);
+  }
+
+  /**
+   * Returns the list of uninstrumented methods. (Note: instrument_jdk() needs to have been called
+   * first.)
+   */
+  public List<String> get_skipped_methods() {
+    return new ArrayList<>(skipped_methods);
```
mernst: I notice that the result of this operation is never modified and is always immediately used, but it is OK to leave the defensive copy in.

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+  }
+
+  /**
+   * Adds a try/catch block around the entire method. If an exception occurs, the tag stack is
```
mernst: Does this method build the whole try/catch block, or just the exception handler?  This documentation is inconsistent with the method name (and the documentation for the return value), it seems.

mernst: "If an exception occurs": does that mean in this method, or at run time for the instrumented code?  I think the latter, but please clarify.

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+   * cleaned up and the exception is rethrown.
+   *
+   * @param mgen method to add exception handler
+   * @return code list for handler, or null if method should not have a handler
```
mernst, line 2088: Where  does the code list go?  At the beginning of the method, the end, or elsewhere?

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+    this.in_jdk = in_jdk;
+    constructor_is_initialized = false;
+    if (Premain.jdk_instrumented) {
+      dcomp_prefix = "java.lang";
+    } else {
+      dcomp_prefix = "daikon.dcomp";
+    }
+    dcomp_marker = ClassDesc.of(Signatures.addPackage(dcomp_prefix, "DCompMarker"));
+    if (BcelUtil.javaVersion == 8) {
+      dcomp_prefix = "daikon.dcomp";
```
mernst: It is confusing that `dcomp_prefix` is used above, then changed, then used again.  Can you give a different name to the two different uses?

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+
+  /**
+   * Creates the local used to store the tag frame and returns it.
+   *
+   * @param mgen method to modify
+   * @param minfo for the given method's code
+   * @return LocalVariable for the tag_frame local
+   */
+  LocalVariable createTagFrameLocal(MethodGen24 mgen, MethodGen24.MInfo24 minfo) {
+    return BcelUtils24.addNewSpecialLocal(mgen, "dcomp_tag_frame$5a", objectArrayCD, minfo, false);
```
mernst: Please document the reason for this specific name.

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+  }
+
+  /**
+   * Generates the code to create the tag frame for this method and store it in tag_frame_local.
```
mernst: What is a "tag frame"?

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+          case Opcode.IFGE:
+          case Opcode.IFGT:
+          case Opcode.IFLE:
+            {
+              return discard_tag_code(inst, 1);
+            }
+
+          // Instanceof pushes either 0 or 1 on the stack depending on whether
+          // the object on top of stack is of the specified type.  We push a
+          // tag for a constant, since nothing is made comparable by this.
```
mernst, line 2337: Is it a *fresh* tag?

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+            return array_store(inst, "iastore", CD_int);
+          case Opcode.LASTORE:
+            return array_store(inst, "lastore", CD_long);
+          case Opcode.SASTORE:
+            return array_store(inst, "sastore", CD_short);
+
+          default:
+            throw new DynCompError("Unexpected instruction opcode: " + inst.opcode());
+        }
+      }
```
mernst, line 2698: This is the end of the code that I suggest abstracting into a separate method.

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+
+      default -> {
+        throw new DynCompError("Unexpected CodeElement: " + ce);
+      }
+    }
+  }
+
+  /**
+   * Adds a call to DCruntime.exit() at each return from the method. This call calculates
+   * comparability on the daikon variables. It is only necessary if we are tracking comparability
```
mernst: What are "the daikon variables"?

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+          li.add(ce);
+        }
+        // skip over the return instruction
+        li.next();
+      }
+    }
+  }
+
+  /**
+   * Return the interface class containing the implementation of the given method. The interfaces of
```
mernst: If this specifically does not return an abstract method (that the interface declares but does not define), please say that.  If it returns any method, then don't use "implementation" here and "defining" in the method name, but use "declaring" instead.

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+      il.add(InvokeInstruction.of(invoke.opcode(), invoke.owner(), nte, invoke.isInterface()));
+      return il;
+
+    } else { // not instrumented, discard the tags before making the call
+      return cleanInvokeTagStack(invoke, classname, returnType, paramTypes);
+    }
+  }
+
+  /**
+   * Process an invoke instruction that calls a non-instrumented method. We need to clean up the tag
```
mernst: "We need to clean up" is ambiguous regarding whether this method does it or the caller must do so.  Please clarify.

The phrasing "process" is vague.  Maybe cut that, and change to "Clean up the tag stack for an invoke instruction...".

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+        System.out.printf("push tag for return type of %s%n", returnType);
+      }
+      il.add(dcr_call("push_const", CD_void, noArgsCD));
+    }
+    il.add(invoke);
+    return il;
+  }
+
+  /**
+   * Returns instructions that will discard any primitive tags corresponding to the specified
```
mernst: Does "discard" mean "pop", or something else?

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+      il.addAll(discard_tag_code(null, primitive_cnt));
+    }
+    return il;
+  }
+
+  /**
+   * Returns true if the invoke target is instrumented.
+   *
+   * @param invoke instruction whose target is to be checked
+   * @param mgen host method of invoke
```
mernst: I'm not sure what "host method" means.  Is it the method that defines the caller?  It it the method that contains the call?

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+              break;
+            }
+            // Recurse looking in the superclass.
+            targetClassname = getSuperclassName(targetClassname);
+          }
+        }
+      }
+    }
+
+    if (op.equals(Opcode.INVOKESPECIAL)) {
```
mernst: Consider moving this to the beginning of the method so that many occurrences above of
```java
             targetInstrumented = false;
             break;
```
can be changed to the simpler `return false;`

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+        }
+        access = Integer_ACC_ANNOTATION;
+      }
+      accessFlags.put(classname, access);
+    }
+    return access;
+  }
+
+  /**
+   * Returns true if the specified classname is instrumented.
```
mernst: "classname" should be "class".
Should "is" be "should be"?  That is, is this method prescriptive or descriptive?  If prescriptive, consider changing to `classShouldBeInstrumented`.  If it's used both ways, document that and don't change the method name.

java/daikon/dcomp/ClassGen24.java
mernst: Enclosing class and superclass are different concepts.  Please make the documentation and name consistent.

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+        poolBuilder.methodRefEntry(
+            runtimeCD, compare_method, MethodTypeDesc.of(CD_boolean, two_objects_arg));
+    il.add(InvokeInstruction.of(Opcode.INVOKESTATIC, mre));
+    il.add(BranchInstruction.of(boolean_if, branch.target()));
+    return il;
+  }
+
+  /**
+   * Handles load and store field instructions. The instructions must be augmented to either push
+   * (load) or pop (store) the tag on the tag stack. This is accomplished by calling the tag get/set
```
mernst: Only for primitives, right?

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+      debugInstrument.log("Adding return local of type %s%n", returnType);
+      minfo.returnLocal =
+          createLocalWithMethodScope(mgen, minfo, "return__$trace2_val", returnType);
+    }
+
+    return minfo.returnLocal;
+  }
+
+  /**
+   * Creates a MethodInfo corresponding to the specified method. The exit locations are filled in,
```
mernst: Please clarify.  I'm not sure what it means to fill in nor what the reflection information is.

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,4916 @@
...
+  }
+
+  /**
+   * Return the fully qualified fieldname of the specified field.
+   *
+   * @param cm class containing the field
+   * @param fm the field
+   * @return string containing the fully qualified name
+   */
+  protected String full_name(ClassModel cm, FieldModel fm) {
```
mernst: I think this method can be static.

mernst (changes requested):

java/daikon/dcomp/BuildJDK24.java
```diff
@@ -0,0 +1,593 @@
...
+    try {
+      // Create the destination directory
+      dest_dir.mkdirs();
+
+      // Process each file.
+      for (String classFileName : class_stream_map.keySet()) {
+        if (verbose) {
+          System.out.println("instrument_classes: " + classFileName);
+        }
+
```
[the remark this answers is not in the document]
markro49: The number of skipped files is currently = 1. Don't this this change is needed.

java/daikon/dcomp/BcelUtils24.java
markro49: renamed and documented

java/daikon/dcomp/BcelUtils24.java
markro49: renamed

java/daikon/dcomp/BcelUtils24.java
markro49: changed to offsetFirstLocalToBeMoved

java/daikon/dcomp/BcelUtils24.java
markro49: improved comments

java/daikon/dcomp/BcelUtils24.java
markro49: improved comments

java/daikon/dcomp/StackMapUtils24.java
markro49: This would be a little messy as both newIndex and newOffset would be computed by the method. So it would need to return a record containing the two.  Is it worth it?

java/daikon/dcomp/StackMapUtils24.java
markro49: so i would make the debugInstrument SimpleLog in StackMapUtils24 public and the insert an assignment to it's 'enable' prior to every call to addNewSpecialLocal?

java/daikon/dcomp/ClassGen24.java
markro49: duplicate

java/daikon/dcomp/CalcStack24.java
markro49: * <p>During this process, it may modify the state of DCInstrument24.locals and DCInstrument24.stacks.

java/daikon/dcomp/CalcStack24.java
markro49: It is a bug in the method being simulated, not in CalcStack. E.g., the method is referencing an element of an uninitialized array.

java/daikon/dcomp/CalcStack24.java
markro49: Traditionally, peek() doesn't modify the stack. I would like to continue to use pop and push.

java/daikon/dcomp/CalcStack24.java
markro49: All of the 'operand stack before:' and 'operand stack after:' comments were taken straight from the The Java Virtual Machine Instruction Set chapter of the Java Virtual Machine Specification.  They use value1, value2, etc.  So I think using the short hand v1, v2, etc. makes sense.  Are there places where I don't follow that convention?

java/daikon/dcomp/OperandStack24.java
```diff
@@ -0,0 +1,237 @@
...
+    return this.clone();
+  }
+
+  /**
+   * @return a hash code value for the object
+   */
+  @Override
+  public int hashCode(@GuardSatisfied OperandStack24 this) {
+    return stack.hashCode();
+  }
```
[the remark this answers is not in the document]
markro49: equals has been modified. all primitives are no longer treated as equivalent. can you suggest a new version of hashCode?

java/daikon/dcomp/DCInstrument24.java
markro49: no, they are unrelated

java/daikon/dcomp/DCInstrument24.java
markro49: This loop copies all the class elements that require no instrumentation from the input class to the output class.  Primarily, fields and attributes.

java/daikon/dcomp/DCInstrument24.java
markro49: You are correct, most of the tests are in Instrument24; however, this particular test needs to inspect the class's CodeModel which is not available until we start the instrumentation process.

java/daikon/dcomp/DCInstrument24.java
markro49: These fields require a CodeBuilder which is not available until we start the instrumentation process.

java/daikon/dcomp/BuildJDK24.java
```diff
@@ -0,0 +1,594 @@
...
+      // Write out the list of all classes in the jar file
+      File jdk_classes_file = new File(dest_dir, "java/lang/jdk_classes.txt");
+      System.out.printf("Writing a list of class names to %s%n", jdk_classes_file);
+      // Class names are written in internal form.
+      try (PrintWriter pw = new PrintWriter(jdk_classes_file, UTF_8.name())) {
+        for (String classFileName : class_stream_map.keySet()) {
+          pw.println(classFileName.replace(".class", ""));
+        }
+      }
+    }
```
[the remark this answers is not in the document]
markro49: solved by not adding these files to class_stream_map

java/daikon/dcomp/BuildJDK24.java
```diff
@@ -0,0 +1,594 @@
...
+      String entryName = path.toString().substring(modulePrefixLength + 1);
+      try {
+        // Get the InputStream for this file
+        InputStream is = Files.newInputStream(path);
+        class_stream_map.put(entryName, is);
+      } catch (Exception e) {
+        throw new Error(e);
+      }
+    }
+  }
```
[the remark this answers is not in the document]
markro49: solved by not adding these files to class_stream_map

java/daikon/dcomp/StackMapUtils24.java
```diff
@@ -0,0 +1,193 @@
...
+   */
+  public static LocalVariable addNewSpecialLocal(
+      MethodGen24 mgen,
+      MethodGen24.MInfo24 minfo,
+      @Identifier String varName,
+      ClassDesc varType,
+      boolean isParam) {
+
+    LocalVariable varNew;
+    // get a copy of the locals before modification
```
mernst: This looks like it's the actual locals, not a copy.
Also, the method side-effects locals and thereby side-effects `mgen.localsTable`.  Please document that (unless I'm mistaken).

java/daikon/dcomp/CalcStack24.java
```diff
@@ -0,0 +1,1014 @@
...
+import java.lang.classfile.instruction.TableSwitchInstruction;
+import java.lang.classfile.instruction.TypeCheckInstruction;
+import java.lang.constant.ClassDesc;
+import java.lang.constant.MethodTypeDesc;
+import java.util.List;
+import java.util.Set;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+/**
+ * This class calculates the state of the operand stack via simulation.
```
mernst, line 55: The class uses both "simulation" and "symbolic execution".  Please choose a term and use it througout.

java/daikon/dcomp/CalcStack24.java
```diff
@@ -0,0 +1,1014 @@
...
+        // Nothing needs to be done.
+        return true;
+      }
+
+      // Technically, a Classfile element, not a PseudoInstruction.
+      case Label l -> {
+        if (DCInstrument24.stacks[instIndex] != null) {
+          // We've seen this label before.
+          DCInstrument24.verifyOperandStackMatches(l, DCInstrument24.stacks[instIndex], stack);
+          // Stacks match; we're done with this worklist.
```
mernst, line 115: What is "this worklist"?

java/daikon/dcomp/CalcStack24.java
```diff
@@ -0,0 +1,1014 @@
...
+
+      // operand stack before: ...
+      // operand stack after:  [empty]
+      case Opcode.ARETURN:
+      case Opcode.DRETURN:
+      case Opcode.FRETURN:
+      case Opcode.IRETURN:
+      case Opcode.LRETURN:
+      case Opcode.RETURN:
+        // execution pump will reset stack
```
mernst, line 246: What is the "execution pump"?

java/daikon/dcomp/CalcStack24.java
```diff
@@ -0,0 +1,1014 @@
...
+      // operand stack before: ..., objectref, [arg1, [arg2 ...]]
+      // operand stack after:  ...
+      case Opcode.INVOKEINTERFACE:
+      case Opcode.INVOKESPECIAL:
+      case Opcode.INVOKEVIRTUAL:
+        stack.pop(); // discard the objectref
+      // may actually be removing an arg, but we'll
+      // account for that when we remove args below
+
+      // fall through is intentional:
```
mernst, line 895: The fallthrough is a bit too clever.  Can you put all 4 `case` labels together, and then within the single body for them all, add an `if (! inst.opcode instanceof INVOKESTATIC)` for the special behavior?

java/daikon/dcomp/CalcStack24.java
```diff
@@ -0,0 +1,1014 @@
...
+      // operand stack before: ..., index
+      // operand stack after:  ...
+      case Opcode.TABLESWITCH:
+        stack.pop(); // discard the index
+        TableSwitchInstruction tsi = (TableSwitchInstruction) inst;
+        addLabelsToWorklist(tsi.defaultTarget(), tsi.cases(), stack);
+        return false;
+
+      // operand stack before: ..., objectref, [arg1, [arg2 ...]]
+      // operand stack after:  ...
```
mernst, line 885: The operand stack after contains the return type, unless the return type is void.

java/daikon/dcomp/CalcStack24.java
```diff
@@ -0,0 +1,1014 @@
...
+      case Opcode.INVOKESPECIAL:
+      case Opcode.INVOKEVIRTUAL:
+        stack.pop(); // discard the objectref
+      // may actually be removing an arg, but we'll
+      // account for that when we remove args below
+
+      // fall through is intentional:
+
+      // operand stack before: ..., [arg1, [arg2 ...]]
+      // operand stack after:  ...
```
mernst, line 898: The operand stack after contains the return type, unless the return type is void.

java/daikon/dcomp/BuildJDK24.java
```diff
@@ -0,0 +1,606 @@
...
+   */
+  public static final String java_home = System.getProperty("java.home");
+
+  /** If true, print information about the classes being instrumented. */
+  private static boolean verbose = false;
+
+  /** Number of class files processed; used for progress display. */
+  private int _numFilesProcessed = 0;
+
+  /** Name of file in output jar containing the static-fields map. */
```
mernst, line 79: Where can I find documentation about the static-fields map?

java/daikon/dcomp/BuildJDK24.java
```diff
@@ -0,0 +1,606 @@
...
+      // Class names are written in internal form.
+      try (PrintWriter pw = new PrintWriter(jdk_classes_file, UTF_8.name())) {
+        for (String classFileName : class_stream_map.keySet()) {
+          pw.println(classFileName.replace(".class", ""));
+        }
+      }
+    }
+
+    // Print out any methods that could not be instrumented
+    print_skipped_methods();
```
mernst, line 205: Should this crash by default if there are any skipped methods, possibly with an override for use during debugging or to work around bugs?

java/daikon/dcomp/BuildJDK24.java
```diff
@@ -0,0 +1,606 @@
...
+          System.out.println("instrument_classes: " + classFileName);
+        }
+
+        if (classFileName.equals("module-info.class")) {
+          System.out.printf("Skipping file %s%n", classFileName);
+          continue;
+        }
+
+        // Handle non-.class files and Object.class.  In JDK 8, copy them unchanged.
+        // For JDK 9+ they have not be added to class_stream_map.
```
mernst: Grammar problem.  Maybe you meant this:

```suggestion
        // For JDK 9+ they have not been added to class_stream_map, so the if condition is never true.
```

java/daikon/dcomp/BuildJDK24.java
```diff
@@ -0,0 +1,606 @@
...
+
+  /**
+   * Instruments the JavaClass {@code jc} (whose name is {@code classFileName}). Writes the
+   * resulting class to its corresponding location in the directory outputDir.
+   *
+   * @param classFile ClassFile of class to be instrumented
+   * @param classModel ClassModel of class to be instrumented
+   * @param loader ClassLoader of class to be instrumented
+   * @param outputDir output directory for instrumented class
+   * @param classFileName name of class to be instrumented (in internal form)
```
mernst, line 535: I am confused about this variable.  Is it the name of a class (which can be in internal form) or is it the name of a file as its name suggests?  The code below does strip ".class" from it.

java/daikon/dcomp/Instrument24.java
```diff
@@ -0,0 +1,374 @@
...
+    // If already instrumented, there is nothing to do.
+    // (This set will be empty if Premain.jdk_instrumented is false.)
+    if (Premain.pre_instrumented.contains(className)) {
+      debug_transform.log("Skipping pre_instrumented JDK class %s%n", binaryClassName);
+      return null;
+    }
+
+    boolean in_jdk = false;
+
+    // Check if class is in JDK
```
mernst, line 181: The following code treats JDK and non-JDK classes rather differently.  I suggest moving, out of the if body, anything that could be relevant to both.  For example, problem classes, $Proxy, and transformers could be moved outside of the if.  Smaller code to compare (between the two branches of the if) will be easier for readers to comprehend.

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,5022 @@
...
+  /** If true, enable operand stack debugging. */
+  protected static final boolean debugOperandStack = false;
+
+  // End of debug loggers.
+
+  /** Keeps track of the methods that were not successfully instrumented. */
+  protected List<String> skipped_methods = new ArrayList<>();
+
+  /** If we're using an instrumented JDK, then "java.lang"; otherwise, "daikon.dcomp". */
+  protected @DotSeparatedIdentifiers String dcompMarkerPrefix;
```
mernst, line 557: `DCInstrument` uses `dcompRuntimePrefix`.  Use the same name here and there.

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,5022 @@
...
+  void skip_method(String m) {
+    skipped_methods.add(m);
+  }
+
+  /**
+   * Returns the list of uninstrumented methods. (Note: instrument_jdk_class() needs to have been
+   * called first.)
+   */
+  public List<String> get_skipped_methods() {
+    return skipped_methods;
```
mernst, line 2079: `DCInstrument wraps this in `new ArrayList<>(...)`.  Which one is right?

java/daikon/dcomp/DCInstrument24.java
```diff
@@ -0,0 +1,5022 @@
...
+    ClassDesc paramTypes[] = mgen.getParameterTypes();
+
+    // Determine the offset of the first argument in the frame
+    int offset = 1;
+    if (mgen.isStatic()) {
+      offset = 0;
+    }
+
+    // allocate an extra slot to save the tag frame depth for debugging
+    int frame_size = mgen.getMaxLocals() + 2;
```
mernst: In `DCInstrument`, this is "... + 1".

mernst (changes requested):

PR #784, opened 2026-04-29
update checker to 3.55.1

markro49:


markro49:
wait until after daikon release

PR #793, opened 2026-06-02
update kvasir tests due to problems discovered using clang

markro49:


mernst:
@markro49 The CI failure is likely due to a change in the Docker images.

PR #809, opened 2026-07-24
fix goal files for fix to array size calculation for multi-dim arrays

markro49:
Merge with https://github.com/codespecs/fjalar/pull/76, which is for the "support-rust" branch.


mernst:
@coderabbitai review

mernst (approved):
LGTM

PR #818, opened 2026-08-04
Create libcalculator.so on host during test

markro49:


markro49:
CodeRabbit complains about a couple of things that I don't think matter - but you're better with Makefiles than me.  The big issue is I don't have access to ARM based computer and I have no idea if the gcc line:
gcc -fPIC -I"${JAVA_HOME}/include" -I"${JAVA_HOME}/include/linux" -shared -o libcalculator.so ncalc.c
is correct for that system.

mernst (approved):
Thanks for the fix!

mernst (approved):
Thanks!

PR #836, opened 2026-09-14
Extract JUnit test class detection into its own method

markro49:
Suggested by Michael Ernst in review of PR #685.

`instrumentClass` in `DCInstrument24` and `instrument` in `DCInstrument` each carried about 150 lines of JUnit test-class detection inline. This moves that into `checkForJunitTestClass` in both instrumenters, leaving one line at the call site.

The long "very tricky special case" comment explaining why JUnit classes need different treatment becomes the method's javadoc, which is where it belongs: it explains why the method exists rather than what the next statement does.

### Behavior change

One, intentional. The `if (junit_test_class) … else …` debug logging used to sit inside the `!in_jdk` guard, so a JDK class logged nothing. It is now outside, so every class logs its result. Debug-only output, enabled by `--debug`.

### Reviewing this

`git diff` renders it poorly, because a ~150-line move with a handful of edits looks like an unrelated delete plus add. Comparing the old in-place block against the new method with indentation and blank lines normalized gives the real change:

```
> private boolean checkForJunitTestClass(@BinaryName String classname) {
> if (in_jdk) {
< if (!in_jdk) {
< // A very tricky special case: ... (10 comment lines, now the javadoc)
> } else {
> }
> return true;
> return false;
< classInfo.isJunitTestClass = junit_test_class;
```

Everything else is byte-identical. The guard is inverted so that the "a JDK class is never a JUnit test class" comment sits where a reader asks the question; the empty `then` branch is deliberate.

The two instrumenters differ in one respect. `DCInstrument24` assigns to `classInfo.isJunitTestClass`, so its call site is that assignment. `DCInstrument` has no such field and consumes `junit_test_class` later in the same method, so its call site keeps the local. The parameter is `@BinaryName` in the 24 port and `@ClassGetName` in the BCEL one, matching each file's existing type for that value.

### Testing

`make compile`, `make JAVA24=1 junit` (107 + 29), `make check-format`, `javadoc -Xdoclint:all`, and all six `typecheck-part1` checkers against the bundled Checker Framework are clean.

No new tests: this is a pure extraction, and the existing suite covers the behavior. I verified nothing was lost by comparing the sorted set of code lines of each whole file before and after; only the structural lines above differ.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01A1De2wQi77Zz4pnapvnFJz


mernst (approved):
Thanks!

PR #837, opened 2026-09-14
Do not consult interfaces when the superclass chain is incomplete

markro49:
Reported by CodeRabbit on PR #830, after that PR had merged. The code it refers to came in with #830's follow-up work, so this is a fix against master rather than a regression from an open branch.

When resolving whether an invoke's target is instrumented, the walk up the superclass chain stops early if a class file cannot be located or read:

```java
if (targetClass == null) {
  // We cannot locate or read the .class file, so the superclass chain ends here.  The
  // method may still be declared by an interface of a class already in the chain.
  if (!isInterfaceMethodInstrumented(chain, methodName, paramTypes)) {
    targetInstrumented = false;
  }
  break;
}
```

Consulting the interfaces of the classes seen so far can report that an instrumented interface declares the method. But the unreadable class may define it concretely, and that class was not instrumented, so the `DCompMarker` overload the caller would then invoke was never generated. The call fails at run time.

Every other misresolution in this method costs precision -- the caller invokes the uninstrumented overload, which always exists, and comparability is lost through that call. This one breaks the call, so it is worth treating differently.

The fix assumes the target is not instrumented, which is what the rest of this resolver does whenever a class file cannot be read.

### Scope

`isInterfaceMethodInstrumented` is called from three places in each instrumenter. The other two run on a **complete** chain -- one after a declaring method is found, one after reaching `Object` -- where consulting the interfaces is correct. Only this one acts on a partial chain, so only this one changes. Applied to both `DCInstrument` and `DCInstrument24`, which are structurally identical here.

### One correction to the report

CodeRabbit predicts `AbstractMethodError`. `NoSuchMethodError` seems likelier: the marker overload does not exist at all, rather than existing without an implementation. Same severity either way.

### Testing

No new test. Triggering this needs `getJavaClass` / `getClassModel` to fail on a mid-chain superclass while an instrumented interface declares the method, which means an unreadable class file -- the existing fixtures have no way to produce one. Constructible with a custom loader, but the scaffolding would be larger than the change; happy to add it if you would like.

`make compile`, `make JAVA24=1 junit` (107 + 29), `make check-format`, `javadoc -Xdoclint:all`, and all six `typecheck-part1` checkers against the bundled Checker Framework are clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01A1De2wQi77Zz4pnapvnFJz


mernst (approved):
Thanks!

